PR #208, opened 2015-01-12
Issue201 durack1 portrait plot template tweaks

doutriaux1:
Issue25 add custom metrics
should be cleaned up now.
portrait_2obs need cleaning up, we should proably wait before approving this one.


doutriaux1:
@durack1 2obs does not seem to be gfdl related. Where should we move it to?

durack1:
@doutriaux1 did you want this merged now? Or did you want to keep working on tweaking 2obs?

durack1:
@doutriaux1 sorry, yeah that was meant to be rewritten and provided as the second test case - so we give them two examples..

durack1:
@doutriaux1 did you also want to commit the *.png result - it would be nice to include this so there is something to compare against and validate they're reproducing the same results..

doutriaux1:
ok so you want to add GFDL results to the 2obs and provide this as an example to them correct?
IF so I will tweak it and add the 2 pngs before we approve this merge ok?

durack1:
@doutriaux1 - yep (provide it to them).. And add example *.pngs

durack1:
@gleckler - take a peek and make sure these are what you want!!??

PR #212, opened 2015-01-14
Issue161 durack1 1p0p0 final tweaks

durack1:
@doutriaux1 these are all updated after running at GFDL, so they are all valid, tested and accurate with paths etc - please merge when you can

Also ran the test case and it returned 6 passes:
```
Ran 6 tests in 77.710s

OK
public1:/home/p1d/150113_metrics/PCMDI_METRICS>
```

@gleckler1 you might want to check the demo/GFDL subdir pngs to make sure they're what you wanted..


gleckler1:
2 obs comparison will be cleaner than 3 obs
p

doutriaux1:
yes but 3 shows off better that we can do ANY number and the code follows not powers of 2.

PR #218, opened 2015-01-21
Issue217 durack1 demo for csiro data

durack1:
@doutriaux1 please take a look and merge ASAP


durack1:
@doutriaux1 actually hang 5 while I also update setup.py to pull these files across during installation..

durack1:
@doutriaux1 she's good to go.. Merge when you can..

doutriaux1:
ok now I'm nervous I don't like you touching setup.py ;)

durack1:
@doutriaux1 hand over the car keys dad..

doutriaux1:
lgtm all these demo starting to be a bit redundant and add a lot of noise we should probably have an option in the install script to turn these on/off or just install one

durack1:
@doutriaux1 well they are hidden in the demo subdir and subdirs beneath them.. The reason I want them in there is that each demo has taken a different path to get the results.. So GFDL required variable remapping, NCAR requires variable remapping and interpolation, CSIRO requires the use of a qsub/quota system to actually use the metrics package.. So I think they're all useful and provide us/me with a quick template to point new users toward..

doutriaux1:
i understand that. nd I'm fine with having them in the repo. Just not quite sure it's worth installing an every users' systems.

durack1:
@doutriaux1 as discussed.. In my head this then provides a ready reference that is available to all users who have successfully installed.. If I was leading a new user through preparing their data and using the package I'd select from one of the 4 test cases, which are all sitting in front of them.. It's easier to point them to the demo subdir than for them to hunt around in the tmp subdirs, or to try and figure out how to get a file from github down to their systems..

I agree if these grow ridiculously we might want to trim things, but I think more examples are more useful than less..

PR #240, opened 2015-02-06
fix #229

### LINKED ISSUE #229: Pass the complete variable name to compute_metrics (ua-XXX)
In case we want to use the variable name within a custom compute_metrics function, it would be useful if 'var' was the same as the variable name provided in the parameter file.
The variable name in the output json file name is var-level, when it is var_level in the parameter file.
Would it be beneficial to have the same naming convention everywhere? (i.e. choose between var-level and var_level)

doutriaux1:
passing var to compute funcs the same way the user def it in parameter file, that allows compute to query the level

### DIFF
--- a/src/python/pcmdi/mean_climate_metrics_calculations.py
+++ b/src/python/pcmdi/mean_climate_metrics_calculations.py
@@ -1,7 +1,9 @@
 import cdms2 as cdms
 import pcmdi_metrics
 
-def compute_metrics(var,dm_glb,do_glb):
+def compute_metrics(Var,dm_glb,do_glb):
+    # Var is sometimes sent with level associated
+    var = Var.split("_")[0]
     cdms.setAutoBounds('on')
     metrics_dictionary = {}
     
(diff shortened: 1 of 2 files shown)


durack1:
@doutriaux1 we need to think about this - so level should use "-" rather than "_" as we've standardised on the output files including "_" as the variable delimiters..

durack1:
@doutriaux1 are you happy to standardise this syntax on `"ua-200"` rather than `"ua_200"` as currently?

doutriaux1:
@durack1 let's talk about it with @taylor13 I'm not excited about '-' maybe some other character like '@' ?

durack1:
@doutriaux1 it seems DRS doesn't define this, yet..

http://cmip-pcmdi.llnl.gov/cmip5/docs/cmip5_data_reference_syntax.pdf

taylor13:
Hi Paul and Charles,

Let's talk on Wed.

Welcome home Paul.

cheers,
Karl

durack1:
So @doutriaux1 @taylor13 did we decide that `"ua@200"` was the path forward here.. @taylor13 does that jive with DRS?

taylor13:
I meant to tell you yesterday.  We have CMIP5 variables named wap500 
(pressure velocity at 500 hPa) and ta700 (air temperature at 700 hPa) 
and there might be some others.

For consistency I recommend leaving out the @.  Of course you could 
include the @ in everything *before* you write the variable and at that 
point eliminate it.

Karl

doutriaux1:
@durack1 so.... what did we decide for this?

durack1:
@doutriaux1 good question - I think something parse-able that we can strip out..

durack1:
@doutriaux1 this is also waiting on me I suppose?

PR #244, opened 2015-02-07
added possibility for custom obs

### LINKED ISSUE #232: Add the possibility to provide a custom dictionary of observations?
The metrics package is already very handy in integrating custom metrics directly from the driver. It would be great if we could also include a custom obs dictionary to complete the list of variables.
For example, I've had requests to compute metrics on LAI, and biogeochemistry fields (chlorophyll and things like that).
It's easily done by adding a couple of lines to the driver:
capture d ecran 2015-01-28 a 15 35 05
Any thoughts?

doutriaux1:
@jservonnat @durack1 
fix #232

### DIFF
--- a/src/python/pcmdi/scripts/pcmdi_metrics_driver.py
+++ b/src/python/pcmdi/scripts/pcmdi_metrics_driver.py
@@ -58,6 +58,12 @@ exec("import %s as parameters" % fnm)
 if pth!="":
     sys.path.pop(-1)
 
+#Checking if we have custom obs to add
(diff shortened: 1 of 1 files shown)


durack1:
@jservonnat - I think this gives you exactly what you wanted..?

jservonnat:
Sounds perfect! Thank you @durack1

durack1:
@doutriaux1 this PR would benefit from some additions to the input_parameter_file_all_options.py file - to expose this new functionality to a user, it also partially solves #253 (which will be resolved with better documentation)

PR #257, opened 2015-04-16
ref obs can now be ref by actual name

### LINKED ISSUE #256: ref from parameter should be able to be defined from actual ref name
not just default or alternate1/2
this as been exposed by issue #255 where alt are not defined and therefore other obs cannot be used although in the file

doutriaux1:
not just default/alternate/etc...
fix #256

### DIFF
--- a/src/python/pcmdi/scripts/pcmdi_metrics_driver.py
+++ b/src/python/pcmdi/scripts/pcmdi_metrics_driver.py
@@ -194,13 +194,17 @@ for var in parameters.vars:   #### CALCULATE METRICS FOR ALL VARIABLES IN vars
         region_name = "%i" % region
       metrics_dictionary["RegionalMasking"][region_name]=region
       for ref in refs:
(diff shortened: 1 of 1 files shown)


src/python/pcmdi/scripts/pcmdi_metrics_driver.py
```diff
@@ -194,13 +194,17 @@ def applyCustomKeys(O,custom_dict,var):
         region_name = "%i" % region
       metrics_dictionary["RegionalMasking"][region_name]=region
       for ref in refs:
-        if ref in ["default","alternate"]:
+        if ref[:9] in ["default","alternate"]:
```
durack1: Is the :9 hard coding a problem?

doutriaux1: it's because we have alternatee alternate1 and alternate2 this covers it all

PR #263, opened 2015-06-05
Fixed reversed order of call to annual_mean.compute()

jkrasting:
We found here at GFDL that the statistics for the annual means had the sign reversed.  I tracked this down to a bug in the order of the the returned objects form annual_mean.compute()

Order should be:
   dm_am, do_am =  pcmdi_metrics.pcmdi.annual_mean.compute(dm,do)


doutriaux1:
@durack1 please update test suite as well next time

PR #265, opened 2015-06-10
Issue219 corrected compute custom only

jservonnat:


jservonnat:
@durack1 @gleckler1 @doutriaux1 here is the new PR (same as my previous PR but corrected, and with an example). Can you tell me if it is correct?
What I've done:
- added the modification to the driver so the user can choose to compute only the PCMDI metrics, only his custom metrics or both
- added a toy custom function to show an example in src/python/pcmdi/example_compute_custom_only.py (and modified __init__.py accordingly)
- added a param file (input_parameters_compute_custom_only.py) to illustrate how we can use this functionnality.
Hoping that this time it will work.

src/python/pcmdi/scripts/pcmdi_metrics_driver.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/data/igcmg/database/PMP_install/PCMDI_METRICS/bin/python
```
durack1: This custom path shebang will need removing - I'd note that you can setup your environment variables so a `which python` will return the correct PMP one by `source ~MetricsInstallDir/bin/setup_runtime.sh or .csh`

jservonnat: I totally agree!
Thank you for this point.

durack1: @jservonnat do you want to submit a revised PR (so just commit the edit/removal of the custom path to your existing branch), that way it's a simple approve, rather than having to clone, edit and then merge it

jservonnat: @durack1 Ok I'll do that.

jservonnat:
@durack1 I made the correction this morning. Tell me if you need anything else.

durack1:
@jservonnat great thanks for that.. Oh you might also want to check your git configuration, apparently you're now a woman named Marie-Alice Foujols according to github https://github.com/PCMDI/pcmdi_metrics/pull/265/commits

jservonnat:
@durack1 @doutriaux1 are you still interested in this functionnality? Today I've updated the MP here at IPSL and looking at the driver I've seen many changes since I made the PR. I don't remember if my PR already had conflicts when I submitted it but today it surely does.
I will add this compute_custom_only functionnality to the new driver for us at IPSL. If you want me to setup a new PR (up-to-date) just tell me.

doutriaux1:
@jservonnat I think it's in already, double check before working too hard on this.

PR #274, opened 2015-07-10
Pep8

doutriaux1:
@durack1 you will like this one it pep8s/flake8s our entire package!
Also added a test to make sure it stays pep8ed


doutriaux1:
that was fast! thanks

durack1:
@doutriaux1 no reason to hold back on good ideas like this.. The quicker they're in, the better off we are!!

PR #302, opened 2016-02-19
299 create dev space

doutriaux1:
@gleckler1 @durack1 let me know if this makes any sense, look at: https://github.com/PCMDI/pcmdi_metrics/blob/299_create_dev_space/doc/HOWTO_devel_package.md

to know how it is supposed to work, I tried to make it very simple and automate the build for you.

### DIFF
--- a/doc/HOWTO_devel_package.md
+++ b/doc/HOWTO_devel_package.md
@@ -1,0 +1,29 @@
+# Adding a Python Devel Package
+
+* create a directory in `src/python/devel` named `YOUR_PACKAGE` (replace with appropriate package name)
+
+* In the above directory create 3 subdirectories:
+ * `lib` where your python packge file will sit
+ * `scripts` where your scripts will be
+ * `data` where files you want to package will be
+
+* you can access your package by import pcmdi_metrics.YOUR_PACKAGE
+
+# `lib` directory
+
+create a `__init__.py` file from which you expose functions you want to access directly from `pcmdi_metrics.YOUR_PACKAGE`
+example:
+`file1.py` in the directory contains a function `foo` rather than access it via `pcmdi_metrics.YOUR_PACKAGE.file1.foo` you can add the follwoing line in your `__init__.py`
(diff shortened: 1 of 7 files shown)


durack1:
@doutriaux1 this is very nifty, but is way over the head of most conventional scientists..

gleckler1:
This should help us get started but its true scientists will struggle.    I don't really get
"lib where **your python packge file** will sit"
using /lib for including functions/modules I get, but what is "your python package _file_"?  How is this different from a "driver script" in the script directory?   Anyhow, thanks for setting this up and we'll start experimenting with it and keep you in the loop as difficulties arise...
FYI a few starters...
/obs_preparation - here Paul and I will develop scripts converting obs monthly data with cmor and climatologies using cmor.  
/variability_santer2011 - here PG will work with Celine on advancing a script she has been developing
/seaice_ivanova2016 - this will be annual cycle sea-ice metrics
... there is going to be an interesting list!

durack1:
@gleckler1 wanna merge and then tweak? Is this what you envisaged?

doutriaux1:
@gleckler1 look into src/python/devel 
There is an example of a `package` called `example_dev`
basically `scripts` (like pcmdi_driver) go in `scripts` (doh..)
all the actual python files to run the code go in `lib` (like files currently in src/python/pcmdi)
any doc/share etc... need go into doc/share subdirectories of `data`

I can walk you through it on Monday.

doutriaux1:
@durack1 let me go through it with @gleckler1 first probably a few tweaks then, then let's merge. Anyway nothing in `devel` will be built unless the user specificcally asks for it, so even if you merge it won't affect end-users.

durack1:
@doutriaux1 no problems.. Both @gleckler1 and I are out east this week, so might be a little harder to get feedback.. I'm aware that we're under a fair time pressure on all this so just want to make sure that I'm/we're not holding up progress by not merging your standing PRs or anything..

PR #324, opened 2016-04-26
WIP Issue 321 enso bellenger

gleckler1:
Jiwoo and Peter still working on this.


durack1:
@lee1043 @gleckler1 if the `durolib` library is becoming useful, we could burying this into the metrics package as a dependency.. It's only a couple of kb in filesize..

PR #337, opened 2016-05-05
disclaimer updated with EOS publication doi

gleckler1:
@doutriaux1 updates disclaimer with EOS paper reference and doi


gleckler1:
@doutriaux1 merge with master ready!

PR #356, opened 2016-05-27
Reorg dict add case id as off param

doutriaux1:
@gleckler1 @durack1 ready for review. It's got @glecker1 std in it and it moves the results in a RESULTS sub sectionn of the dict.


gleckler1:
@doutriaux1 @durack1 So far working as advertised... crunchin away ... thks!

durack1:
@gleckler1 @doutriaux1 v1.1 here we come!

PR #365, opened 2016-06-24
removed some param files from doc

gleckler1:
@doutriaux1 @durack1  starting to do some clean up


durack1:
@doutriaux1 I'll leave this purge up to you.. You're more aware than me what options are currently available in the param files and what/isn't included in the demo files..

PR #366, opened 2016-06-27
updated obs

gleckler1:
@doutriaux1   This is for updating the obs json files.   Note identical versions of this file reside in /doc and src/python/pcmdi/scripts ... of course 2 copies are not needed but we've agreed to keep one in /doc.   The second could either be removed from  src/python/pcmdi/scripts file could be removed or made as a symbolic link.


gleckler1:
@doutriaux1 @durack1 not sure why this is not going anywhere... should PG just get in habbit of skipping pull request and go straight to master merge?

durack1:
@doutriaux1 @gleckler1 it seems that we have a build/test failure here: https://travis-ci.org/PCMDI/pcmdi_metrics/builds/140651454#L1012-L1031

gleckler1:
Still going solo...

PR #383, opened 2016-07-19
a few checks to make sure parameter file defines necessary things

doutriaux1:
@durack1 a quicky


durack1:
@doutriaux1 perfect.. Will wait for the `travis` green light (on linux only..) and then merge

PR #386, opened 2016-07-19
343 pjg simple demos

doutriaux1:


doutriaux1:
@gleckler1 please do not merge yet. I'm cleaning it up

demo/parameter_files/pmp_input_parameters_demo1.py
```diff
@@ -1,8 +1,11 @@
 ################################################################################
-#  SAMPLE INPUT PARAMETER FILE FOR THE PCMDI METRICS PACKAGE (PMP V1.1)
+#  THIS IS A DEMO INPUT PARAMETER FILE FOR THE PCMDI METRICS PACKAGE (PMP V1.1)
 #  THIS IS A VERY SIMPLE EXAMPLE, ONLY COMPUTING STATISTICS FOR ONE MODEL VERSION AND ONE VARIABLE 
 #  SAMPLE MODEL AND OBSERVATIONAL DATA TO RUN THIS CODE ARE AVAILABLE FROM: 
-# 
+#  http://oceanonly.llnl.gov/gleckler1/pmp-demo-data/pmpv1.1_demodata.tar
```
doutriaux1: @gleckler1 I think the demo should do this for them, or at least try to. Do you want me to set this up?

demo/parameter_files/pmp_input_parameters_demo1.py
```diff
@@ -14,15 +17,17 @@
 ###############################################################################
 ## DATA LOCATION: MODELS, OBS AND METRICS OUTPUT
 ## ROOT PATH FOR MODELS CLIMATOLOGIES
+## THE MODEL AND OBSERVATIONAL DATA USED FOR THIS TEST CAN BE DOWNLOADED FROM:
+
 #mod_data_path = '/work/metricspackage/mod_clims/cmip5-amip'
-mod_data_path = './demo_obs_model_data/mods/'
-filename_template = "pr_Amon_ACCESS1-0_amip_r1i1p1_197901-198912-clim-ac.nc"
+mod_data_path = './cmip5clims_metrics_package-amip/'
```
doutriaux1: same here I can set it up to be some arbitrary path

demo/parameter_files/pmp_input_parameters_demo1.py
```diff
@@ -14,15 +17,17 @@
...
+filename_template = "pr_ACCESS1-0_Amon_amip_r1i1p1_198001-200512-clim.nc"
```
doutriaux1: yuk, that does not show them how to make it general.

demo/parameter_files/pmp_input_parameters_demo1.py
```diff
@@ -40,7 +45,7 @@
 period = '1979-1989'  # PERIOD OF CLIMATOLOGY
 realization = 'r1i1p1' # REALIZATION
 
-
+save_mod_clims = False
```
doutriaux1: not needed anymore, but nice to show them that they can turn it on. We should add a comment on top of it.

demo/parameter_files/pmp_input_parameters_demo1.py
gleckler1: If you want to improve the demo that would be really great.  But 3 things to keep in mind: 1) think python newbie, 2) keep as simple as possible, and 3) real documentation  ;-)

setup.py
```diff
@@ -46,20 +46,16 @@
...
-demo_ACME_files = glob.glob("demo/ACME/*.py")
-demo_CSIRO_files = glob.glob("demo/CSIRO/*.py")
-demo_CSIRO_files += glob.glob("demo/CSIRO/*.png")
-demo_CSIRO_files += glob.glob("demo/CSIRO/*.sh")
-demo_GFDL_files = glob.glob("demo/GFDL/*.py")
-demo_GFDL_files += glob.glob("demo/GFDL/*.png")
-demo_NCAR_files = glob.glob("demo/NCAR/*.py")
-param_files = glob.glob("doc/parameter_files/*.py")
-
+demo_ACME_files = glob.glob("demo/external-setups/ACME/*.py")
```
doutriaux1: and why do we need to package these files? These are examples/docs we can link to these from some wiki. If yuor demo does not use it, then it shouldn't be packaged. No-one will look at it buried inside the python distribution.

gleckler1: The json files in the demo directory will be used for various demo scripts

durack1:
@doutriaux1 looks like `travis` is reporting errors..

PR #387, opened 2016-07-19
384 pjg plot json query

gleckler1:
@doutriaux1   This is garbled with the merged to master that needed to be undone, probably because PG mucked up setup.py.   But the examples that are introduced ....


gleckler1:
@doutriaux1 pretty garbled... two examples here.. one is read json the other is sample portrait..

durack1:
@gleckler1 it seems you have to merge with master here (two files are conflicting see above) and then this should be good to merge (once the test passes)

doutriaux1:
the basic_portrait.py file should be updated to use the JSON class.

durack1:
@doutriaux1 curious about this one too?

PR #399, opened 2016-08-23
added new zonal bands

gleckler1:
@doutriaux1 @doutriaux1  need to add this quickly to advance a request by ACME team


doutriaux1:
Can't we get better names?

gleckler1:
These are very explicit... more so than NHEX, etc.   Also this is the lingo used in science papers.   We could remove the "-" if you prefer, e.g.,  have "20S20N" rather than "20S-20N"

doutriaux1:
@gleckler1 yes I think it's better with the `-` sign, these can always end up confusing the software. While we are at it, should we be consistent and rename the `NHEX`, etc.. ones?

gleckler1:
@doutriaux1  Lets leave NHEX, TROPICS, etc for a little while longer while be prepare for downstream implications (plotting etc).   All "-" have been removed from region names so this should be ready to push :+1:

doutriaux1:
@gleckler1 you do realize these regions are now just here ready to be consumed but not in the default list, right?

gleckler1:
@doutriaux1  Its ok that they are not part of the default list.   They are activated easy enough by setting the "regions" in the parameter file.   The default list will be revisited in the months ahead as we merge other branches currently under development.

PR #401, opened 2016-08-23
399 pjg add area weighting

gleckler1:


src/python/pcmdi/mean_climate_metrics_calculations.py
```diff
@@ -43,7 +43,7 @@ def compute_metrics(Var, dm, do):
 
     # SET CONDITIONAL ON INPUT VARIABLE
     if var == 'pr':
-        conv = 1.e5
+        conv = 86400.
```
durack1: @gleckler1 this could also be true of `wfo` and other vars..?

src/python/pcmdi/meanabs_xy.py
```diff
@@ -1,5 +1,5 @@
 import MV2 as MV
```
durack1: @gleckler1 this import is inconsistent with the `src/python/pcmdi/bias.py` above - `import MV2` no `as MV`

gleckler1:
@doutriaux1 @durack1 @lee1043 @golaz @covey This should fix the missing area-weights in bias and MAE calculations by using cdutil.averager instead of MV.averager.   All of these should be independently verified as a baseline.    To deal with moving target issues (platform dependent precision, changes in ESMF regridding, etc.) we should strive to document as all workflow (version PMP/UV-CDAT, platform, etc) in the json files, and, as part of the test suite document results and compare with results from other platforms, software versions, etc.

gleckler1:
@durack1 MV2 no used consistently across statistics

doutriaux1:
@gleckler1 you will need to update the tests as well if you change the maths...

src/python/pcmdi/mean_climate_metrics_calculations.py
doutriaux1: @durack1 @gleckler1 this is due to some `bad units`  we should probably enforce S.I units as a requirement and drop this.

src/python/pcmdi/meanabs_xy.py
doutriaux1: I agree with @durack1 we should not use `import blah as b` anywhere let's just use the real module names as much as possible, others might be reading this source code you might as well help them learning our tools the proper way. Beside this is 17 add characters, I don't think we use `MV2` 17 times in the file so it's a net loss in typing :wink:

src/python/pcmdi/meanabs_xy.py
```diff
@@ -10,5 +10,8 @@ def compute(dm, do):
             "Absolute Difference Between Model And Observation",
             "Contact": "Peter Gleckler <gleckler1@llnl.gov>",
         }
-    mae = MV.average(MV.absolute(MV.subtract(dm, do)))
+    absdif = MV.absolute(MV.subtract(dm, do))
+    mae = cdutil.averager(absdif,axis='xy',weights='weighted')
```
doutriaux1: Are you positive we only send `xy` data to this? never any `tz` ? `weights="weighted"` is the default so probably not needed.

gleckler1:
@doutriaux1 Currently we do not have a xyt bias or MAE error but we likely will at some point so this branch will be updated this morning to include "xy" or "xyt" so that it is more consistent with RMS and corr.

gleckler1:
@doutriaux1 @durack1  To be consistent with other statistics, as suggested bias_xyt and mae_xyt have been added although not currently used to produce statistics in json files.

A longstanding lingering but subtle concern involves time averaging with climatologies.  It is presumed that this is done properly with cdutil.averager('xyt') - CD, this should be verified.   

Seasonal means are currently hardwired by index https://github.com/PCMDI/pcmdi_metrics/blob/master/src/python/pcmdi/seasonal_mean.py   ... one day when we finally get cmor producing time bounds on climatolgies correct we can do this more properly.

src/python/pcmdi/bias_xyt.py
```diff
@@ -0,0 +1,14 @@
+import MV2
+import cdutil
+
+def compute(dm, do):
+    """ Computes bias"""
+    if dm is None and do is None:  # just want the doc
+        return {
+            "Name": "Bias",
+            "Abstract": "Compute Full Average of Model - Observation",
+            "Contact": "Peter Gleckler <gleckler1@llnl.gov>",
```
doutriaux1: want to convert to pcmdi_metrics@llnl.gov ?

src/python/pcmdi/bias_xy.py
```diff
@@ -9,4 +9,6 @@ def compute(dm, do):
             "Abstract": "Compute Full Average of Model - Observation",
             "Contact": "Peter Gleckler <gleckler1@llnl.gov>",
```
doutriaux1: convert email here as well?

src/python/pcmdi/bias_xyt.py
durack1: @doutriaux1 @gleckler1 it might be worth doing a global replace throughout the whole repo?

gleckler1:
@doutriaux1 @durack1  yes, email should be updated consistently in all files.   OK with calc changes?  Test suite needs to be updated by either CD or PD.... need to push this quickly for use by local team

doutriaux1:
@gleckler1 @durack1 you guys need to learn how to update the test suite, I will show you.

gleckler1:
@doutriaux1 @durack1  TEST results now updated

durack1:
@gleckler1 @doutriaux1 it seems at least in the first test, it's only the `portrait` that's failing:
```
======================================================================
ERROR: test_portrait (test_portrait.TestGraphics)
  File "test/graphics/test_portrait.py", line 216, in test_portrait
    sys.exit(ret)
SystemExit: -1
Ran 10 tests in 139.909s
FAILED (errors=1)
```

gleckler1:
@doutriaux1 @durack1  its graphics in both tests

durack1:
@doutriaux1 this is a `vcs` palette issue, maybe so over to you..

doutriaux1:
@durack1 I suspected that. vcs changed it's default palette. I'll fix it so that it will not be palette dependent.

gleckler1:
@doutriaux1 @durack1  additional statistic now in there... should we just merge this with master even though graphics test fail?   Turn of graphics test?

durack1:
@gleckler1 @doutriaux1 I'd recommend that the problem is solved, not turning stuff off.. But the time delay in resolving these kind of issues is certainly causing problems

durack1:
@doutriaux1 we need some guidance on this, and solving the graphic palette issue.. I've seen that the following palettes are available in `2.6`:
```
In [7]: x.listelements('colormap')
Out[7]: 
['AMIP',
 'NCAR',
 'bl_to_darkred',
 'bl_to_drkorang',
 'blends',
 'blue_to_grey',
 'blue_to_grn',
 'blue_to_orange',
 'blue_to_orgred',
 'brown_to_blue',
 'categorical',
 'default',
 'grn_to_magenta',
 'ltbl_to_drkbl',
 'rainbow',
 'rainbow_no_grn',
 'sequential',
 'white_to_blue',
 'white_to_green',
 'white_to_magenta',
 'white_to_red',
 'white_to_yellow']
```
And we should be able to solve this issue by adding
```
In [8]: x.setcolormap('rainbow')
```
To `test_portrait.py`, regenerating the test png and voila? Correct?

doutriaux1:
@durack1 I don't think it's the issue (it might be) I think the issue is that we set some colors via their numbers (240 to 255 `used` to be reserved) but not anymore since we can now set colors via strings (`"blue"` , `"red"`, etc...) or `(r,g,b)` values. So I just need to substitute the old color assignment from index to actual value.

durack1:
@doutriaux1 well that is certainly over my head.. I was trying to find documentation as to what palettes/colormaps are shipped in `2.6.1` and couldn't find one..

doutriaux1:
in the upcoming doc we have a section on this.

gleckler1:
@doutriaux1   We need this merged into master and updated in conda NOW - PLEASE!   Better if graphics test suite passes but if not, its much more important that we make this updated capability available now.

src/python/pcmdi/meanabs_xyt.py
```diff
@@ -0,0 +1,17 @@
...
+    """ Computes Mean Absolute Error"""
+    if dm is None and do is None:  # just want the doc
+        return {
+            "Name": "Mean Absolute Error",
+            "Abstract": "Compute Full Average of " +
+            "Absolute Difference Between Model And Observation",
+            "Contact": "pcmdi-metrics@llnl.gov",
+        }
+    absdif = MV2.absolute(MV2.subtract(dm, do))
+    mae = cdutil.averager(absdif,axis='xyt',weights='weighted')
```
doutriaux1: @gleckler1 there is a meanabsdiff in genutil.statistics and it has an option `centered=1` do we care about this here or not?

durack1: @doutriaux1 good question, are the results very different?

src/python/pcmdi/mean_climate_metrics_calculations.py
durack1: @doutriaux1 can some udunits magic solve this?

doutriaux1: @durack1 no you can't ask udunits to go from mm/day to kg/m2/s :wink:

gleckler1: @doutriaux1 @durack1   .... original units where kg/m2/s which are MKS... added conversion because a) mm/day are more commonly used for interpretation and b) kg/m2/s required many more significant digits than all vars

durack1: @doutriaux1 can we add this conversion to the salinity kludge Ina branch somewhere?

PR #410, opened 2016-09-14
fixed cor_xy_an for precip by removing conv

gleckler1:
@doutriaux1  oops... missed a push....


doutriaux1:
@gleckler1 probably need to update tests 😉

PR #417, opened 2016-09-15
Json class

doutriaux1:


doutriaux1:
@durack1 please review it fixes all tests as well.

src/python/pcmdi/io.py
```diff
@@ -31,3 +38,249 @@ def __init__(self, root, var, obs_dic, reference="default", file_mask_template=N
...
+                            areal = aref[real]
+                            areal2 = {"source": src}
+                            for region in areal.keys():
+                                # print "\t\t\tREGION:",region
+                                reg = areal[region]
+                                if region == "global":
+                                    region2 = ""
+                                else:
+                                    region2 = region + "_"
+                                areal2[region2 + "global"] = {}
```
durack1: @doutriaux1 should these be hard-coded? I think using the region list (which is growing) might be a better path forward see share/default_regions.py

doutriaux1: no need to, this is the converter from the older (deprecated) json format, which was not using this dictionary.

PR #442, opened 2016-10-25
fixed monsoon keys redundant with variability modes

gleckler1:
@durack1  discovered redundancy in dict keys, now fixed.   Ready to go....


durack1:
@gleckler1 all those green ticks..

PR #448, opened 2016-11-14
trying to use latest uvcdat with conda-forge channel

doutriaux1:


durack1:
@doutriaux1 @gleckler1 magic! tests pass

PR #452, opened 2016-11-29
Travis fix cdp

zshaheen:


durack1:
@zshaheen not sure it's a good idea to remove all the tests

@doutriaux1 @gleckler1

zshaheen:
@durack1 The tests are still in test root directory. These were copies in src/test/.

PR #453, opened 2016-11-30
Pull request cdp

zshaheen:


durack1:
@zshaheen your commit history tells quite the colourful story..

gleckler1:
@zshaheen  Could we just merge the parser for now?

src/python/io/pmp_io.py
```diff
@@ -395,7 +489,9 @@ def addJson(self, filename):
                     varnm += "-%i" % int(var["level"] / 100.)
             tmp_dict = {varnm: tmp_dict["RESULTS"]}
         else:
-            tmp_dict = tmp_dict["RESULTS"]
+            tmp_dict = tmp_dict["RESULTS
```
durack1: @zshaheen you're missing a ']' here

zshaheen: Yeah, I noticed that. Thanks for the heads up.

zshaheen:
@doutriaux1 I just called the base module pmp_io, since all of the other new pmp/cdp/acme models are prefixed as such. Doesn't really matter, so I'll revert it to base.

Also the new driver works with the demo, but I was hesitant to just push the new stuff on the users and completely remove any reference to the old one. If you're cool with it, then it's fine.

I'll fix the stuff you talked about this week and will let you know.

doutriaux1:
@zshaheen yes let's do a complete switch off.

zshaheen:
@doutriaux1 Think this is everything you wanted. Let me know if you want more changes.

PR #477, opened 2017-03-06
pcmdi_metrics_driver.py dealing with ocean vertical levels

jservonnat:
Version of the pcmdi_metrics_driver.py that deals with ocean vertical levels (same mechanism as for the atmosphere); requires to and so pre-computed on standard vertical levels.
I've made a separate driver since I haven't tested the new version yet (demo failing).

### DIFF
--- a/src/python/pcmdi/scripts/pcmdi_metrics_driver_OceanLevels.py
+++ b/src/python/pcmdi/scripts/pcmdi_metrics_driver_OceanLevels.py
@@ -1,0 +1,757 @@
+#!/home/jservon/anaconda2/envs/PMP_nightly/bin/python
+#
+#
+#  USER INPUT IS SET IN FILE "input_parameters.py"
+#  Identified via --parameters key at startup
+#
+#
+import pcmdi_metrics
+import sys
+import argparse
+import os
+import json
+import genutil
+import warnings
+import cdms2
+import MV2
+import cdutil
+import collections
+import cdat_info
+import unidata
+
+# Statistical tracker
+cdat_info.pingPCMDIdb("pcmdi_metrics", "pcmdi_metrics_driver")
+
+# Before we do anything else we need to create some units
+# Salinity Units
+unidata.udunits_wrap.init()
+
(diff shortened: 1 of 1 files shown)


src/python/pcmdi/scripts/pcmdi_metrics_driver_OceanLevels.py
```diff
@@ -0,0 +1,757 @@
...
+
+# Statistical tracker
+cdat_info.pingPCMDIdb("pcmdi_metrics", "pcmdi_metrics_driver")
+
+# Before we do anything else we need to create some units
+# Salinity Units
+unidata.udunits_wrap.init()
+
+# Create a dimensionless units named dimless
+unidata.addDimensionlessUnit("dimless")
```
durack1: Nice, we were going about this another way.. But this looks promising @dnadeau4 @doutriaux1

durack1:
@jservonnat thanks for submitting this.. We'll need to get the test passing before considering merging (not sure what the issue is). We'll also have to figure out how best to integrate your contribution, I'm not sure that another (separate) driver file is the best path forward.. But we'll have to think about this as it's almost 1000 lines of code..

durack1:
@jservonnat the error could just be because you haven't `flake8-ed` your file, which is a trivial one line command (it cleans up the formatting) @doutriaux1 is the implementer of this, and he can defend his stance:
```
======================================================================
FAIL: flake8 (test_flake8.TestFlake8)
  File "/home/travis/build/PCMDI/pcmdi_metrics/test/test_flake8.py", line 36, in flake8
    self.assertTrue(out == "")
AssertionError: False is not true
Ran 13 tests in 277.939s
FAILED (failures=1)
```

doutriaux1:
@gleckler1 @jservonnat is this still something we want?

doutriaux1:
@jservonnat @gleckler1 what do we do about this one?

gleckler1:
@doutriaux1  We can close for now.  When we get around to this we will do it differently.

PR #487, opened 2017-03-28
example results removed

gleckler1:
These files need to be removed from master


gleckler1:
@doutriaux1 @durack1  This one should be trivial.  If its not merged by somebody else COB today I will do it.

doutriaux1:
done

Issue #303, opened 2016-02-20
Resolve precision issues in test suite results


durack1:
This issue was highlighted in https://github.com/PCMDI/pcmdi_metrics/issues/279#issuecomment-171009552, however is reproduced here to simplify the multiple problems covered in #279 
```
@jservonnat @doutriaux1 ok so we have:
metric: repo # | test #
"bias_xy_djf_GLB": "0.29" | 0.30
"bias_xy_djf_SHEX": "2.26" | 2.27
"bias_xy_jja_SHEX": "0.70" | 0.69
"bias_xy_mam_GLB": "0.13" | 0.14
"bias_xy_mam_NHEX": "-0.95" | -0.94
"bias_xy_son_NHEX": "-1.18" | -1.19
"bias_xy_son_SHEX": "0.67" | 0.66
"mae_xy_djf_NHEX": "1.30" | 1.29
"mae_xy_jja_NHEX": "1.11" | 1.12
"mae_xy_mam_NHEX": "1.18" | 1.17
...
```

The resolution requires a more intelligent (precision aware) dealing of result comparison in the test suite. @doutriaux1 has increased precision to 3 places in #301, but this will likely need a further tweak to resolve the issue for good.

durack1:
@doutriaux1 is this a dupe of #380? If so I suggest we close one of these dupes

doutriaux1:
yes and no, this defintely holds #380

gleckler1:
@doutriaux1 @durack1  What is the status of this - should it be merged with another issue?

durack1:
@gleckler1 my understanding is that this requires more work and is not resolved yet - [we're only testing against `linux` in the travis tests](https://github.com/PCMDI/pcmdi_metrics/blob/master/.travis.yml#L1-L2) - this needs to be solved so that `OS X` tests can be turned back on

durack1:
@zshaheen it would be great to entrain you on getting to the bottom of this issue

durack1:
@gleckler1 @doutriaux1 really, we've fixed this? If yes that's a huge step forward and we should push for a new release ASAP

PR #490, opened 2017-03-31
newer results

doutriaux1:
@gleckler1 @durack1 @dnadeau4 this is a bit puzzling

These are the diff using esmpy. Denis fixed a problem with masking in this version but it's dreadfully slow (52mn for the test suite to run). What worries me is that the results are very different. We need to take a careful look. Please do NOT merge this.


doutriaux1:
I do want you guys to take a look and comment please!

doutriaux1:
The pmp code is identical only cdms2 changed.

gleckler1:
Lets meet Tuesday to discuss how we compare the versions and quantify %wise how the results differ.    We need to understand the discrepancies.

dnadeau4:
is it still slow with the latest way to mask data in CDMS?

dnadeau4:
We could go on webex today,  I am here until 6:30pm.

doutriaux1:
The good news is that Denis' new fast version give similar/identical results as the slow one. I think we are good and that it was due to the esmf misinterpolation.

test/pcmdi/customRegions/tas_2.5x2.5_regrid2_linear_metrics.json
```diff
@@ -806,6 +806,23 @@
...
+        "platform": {
+            "OS": "Linux",
+            "Version": "4.8.0-41-generic",
+            "Name": "drdoom"
+        },
+        "userId": "doutriaux1",
+        "osAccess": false,
+        "commandLine": "/home/doutriaux1/anaconda2/envs/test_nightly/bin/pcmdi_metrics_driver.py -p test/pcmdi/region_specs_test.py",
+        "date": "2017-03-31 12:05:11",
+        "conda": {},
```
durack1: @doutriaux1 looks like this provenance stuff is no propagating..

doutriaux1:
@durack1 the provenance is not the issue here, it's ignored by the terst suite, the issue is the number difference

durack1:
@doutriaux1 got you, it could be useful to generate some diff maps to try and understand this issue.. At the same time it could be a good idea to determine the `OS X` vs `linux` differences #303 - they may be related

gleckler1:
@doutriaux1  Not up to speed on this one.   Lets talk about it.  Generating difference maps definitely a good idea.

PR #494, opened 2017-04-11
New2 mac

doutriaux1:


doutriaux1:
these are generated on a mac and compare to the pr #493

PR #495, opened 2017-05-12
edit travis

doutriaux1:


doutriaux1:
@zshaheen seems like we still have differences between mac and linux

zshaheen:
That sucks.

The tests that failed in tests for this PR were gensftlf_test and salinity_test. However, as seen [here](https://www.diffchecker.com/xvMUvz5P), both of these tests (and all of them actually) had the same errors on mac as linux.

doutriaux1:
it passes on Mac here but not on Linux.

zshaheen:
@doutriaux1 I guess I'll run the code in this PR on the same linux vm and mac. I'll see if the results match travis and hopefully will get back to you next week.

doutriaux1:
thx @zshaheen. Look at the .travis.yml to be sure to get the right environment (2.10).

zshaheen:
Okay, I'll make sure. Thanks again for doing this too.

doutriaux1:
@zshaheen @gleckler1 @durack1 @lee1043 seems like the Linux vs /MAc diff are on the 3rd digit after the comma. So it may just be by comparison tool that is not respecting the epsilon difference any longer. There's hope.

doutriaux1:
ok there are a few lines above the threshold of .01
```
comparing: /export/doutriaux1/git/pcmdi_metrics/pcmdi_install_test_results/metrics_results/gensftlfTest/tas_2.5x2.5_esmf_linear_metrics.json /export/doutriaux1/git/pcmdi_metrics/test/pcmdi/gensftlfTest/tas_2.5x2.5_esmf_linear_metrics.json
Failing line: -                            "ann": "290.102", (we read: 290.134 ) diff:  0.032 tolerance: .01
Failing line: -                            "son": "289.779", (we read: 289.811 ) diff:  0.032 tolerance: .01
Failing line: -                            "mam": "290.368", (we read: 290.402 ) diff:  0.034 tolerance: .01
Failing line: -                            "jja": "289.906", (we read: 289.943 ) diff:  0.037 tolerance: .01
Failing line: -                            "djf": "290.364" (we read: 290.38 ) diff:  0.02 tolerance: .01
Failing line: -                            "ann": "3.469" (we read: 3.54 ) diff:  0.08 tolerance: .01
Failing line: -                            "ann": "10.361", (we read: 10.31 ) diff:  -0.051 tolerance: .01
Failing line: -                            "son": "10.648", (we read: 10.594 ) diff:  -0.054 tolerance: .01
Failing line: -                            "mam": "10.557", (we read: 10.501 ) diff:  -0.056 tolerance: .01
Failing line: -                            "jja": "10.674", (we read: 10.593 ) diff:  -0.081 tolerance: .01
Failing line: -                            "djf": "10.497" (we read: 10.47 ) diff:  -0.02 tolerance: .01
Failing line: -                            "ann": "10.656" (we read: 10.6 ) diff:  -0.05 tolerance: .01
```

doutriaux1:
The question really is that I should probably not do a straight diff but a percentage diff? If so what threshold do you recommend?

durack1:
@doutriaux1 is the regridding identical for reach? This is way larger than a precision error, so there must be something else that we're missing

doutriaux1:
Same code on both… The thing that is worrisome is that cdms regrid test pass on mac and linux, so they must give identical results. right @nadeau4 ?

durack1:
@dnadeau4 you have a question above

doutriaux1:
the other failing test'sdiff:
```
comparing: /export/doutriaux1/git/pcmdi_metrics/pcmdi_install_test_results/metrics_results/salinityTest/sos_2.5x2.5_esmf_linear_metrics.json /export/doutriaux1/git/pcmdi_metrics/test/pcmdi/salinityTest/sos_2.5x2.5_esmf_linear_metrics.json
Failing line: -                            "djf": "0.893" (we read: 0.87 ) diff:  -0.02 tolerance: .01
Failing line: -                            "djf": "1.586" (we read: 1.56 ) diff:  -0.02 tolerance: .01
```

dnadeau4:
Actually, some test I assert equal, but some tests I assert 2 digits after the decimal points.  

```
self.assertEqual(ESMF3x4[0, 0], 0.5)
...
self.assertAlmostEqual(s3.max(), 66.99, places=2)
```
Denis

lee1043:
If DJF and ANN has larger difference than others, one contributor could be updating cdutil to consider leap year bound which @dnadeau4 helped.

doutriaux1:
@lee1043 good idea I doubt it because both are using the **EXACT** same version of everything. And the leap year thing is fixed in cdutil now.

doutriaux1:
@dnadeau4 that could be it. The salinity test is definitely using esmpy. the gensftlf might be too while generating the sftlf I'm not sure which regridTool it's using by default. I will check.

@zshaheen on Monday let's work on this and save the climos from these two test and see where the diffs are.

PR #522, opened 2017-12-16
Adds an example to generate a portrait plot with modal support

James-Crean:


durack1:
@doutriaux1 @James-Crean this is great! Do we have a live demo somewhere to see how things look?

PR #526, opened 2018-02-16
Py3

doutriaux1:
All done. Py3 fails because it runs with esmf 7.1 which is different form 7.0 (py2 fails both from master or this with esmf 7.1)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ script:
   - git clean -fd
   - source activate py3
   - python setup.py install
-  - python run_tests.py -v2
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then python run_tests.py -v2 ; fi
   - source deactivate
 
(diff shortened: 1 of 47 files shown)


durack1:
@doutriaux1 time for a new baseline with ESMF7.1 (if we get comparable results using `regrid2` for the atmos grids)

@gleckler1

doutriaux1:
@gleckler1 thiking of switching the tests to use linear at least until https://github.com/UV-CDAT/cdms/issues/231 is fixed. What do you think?

doutriaux1:
@durack1 agreed, that's why we could also disable conservative completely at the moment and raise an error, until this bug is addressed. Thoughts?

durack1:
@doutriaux1 this is not released, right? A warning would be a good idea until the behaviour is acceptable

gleckler1:
@doutriaux1 @dnadeau4 @durack1   Seems like a good thing to do if not too onerous.  I have told the E3SM people using the PMP to not use ESMF conservative until we resolve this, but maybe better is to disable for now.

doutriaux1:
ok will disable on Monday and merge this in master.

PR #537, opened 2018-04-27
fixed glob prob

gleckler1:
@doutriaux1  Finally got diurnal -p working again... it was not a py print problem nor a glob prob, but it appears as thought the aliases are not working for cia...  anyhow, this now works


doutriaux1:
@gleckler1 yes in parameter file aliases are not to be used.

doutriaux1:
@gleckler1 so we can close this now? Correct?

doutriaux1:
@gleckler1 I think we can close this, right?

gleckler1:
@doutriaux1 this basically stems from param file not working with diurnal, so when that gets fixed this goes...

PR #561, opened 2018-07-19
fixed parallelize_driver, changed climo to take model and filename te…

doutriaux1:
…mplate as input rather than input file


doutriaux1:
@gleckler1 please try this branch. I updated the example param file as well.

gleckler1:
@doutriaux1  Where is updated example param file?

doutriaux1:
in examples/parameter_files

gleckler1:
@doutriaux1 does not look like it handles models.   I'm off until z AM

doutriaux1:
@gleckler1 @durack1 please review the just added documentation. Thanks.

gleckler1:
@doutriaux1 will test again now.

gleckler1:
@doutriaux1  I've not been able to get it to work.  It seems to want to "open" the driver, not execute it.   I wonder if part of the problem is that we can' have a perfect template for the model data because of the different version numbers on the xmls (e.g., ...Amon.rlut.ver-**v20130401**.latestX.xml)
parallelize_driver.py --driver pcmdi_compute_climatologies.py -p cmip_clims_param_parallel.py

/export/gleckler1/anaconda2/envs/pmp_0717/bin/python pcmdi_compute_climatologies.py -p /tmp/tmpSxXLOY.pyExecuting:
 /export/gleckler1/anaconda2/envs/pmp_0717/bin/python pcmdi_compute_climatologies.py -p /tmp/tmpf66R5R.py
True
True
Executing: /export/gleckler1/anaconda2/envs/pmp_0717/bin/python pcmdi_compute_climatologies.py -p /tmp/tmp0C5Zf_.py
Executing:True
 /export/gleckler1/anaconda2/envs/pmp_0717/bin/python pcmdi_compute_climatologies.py -p /tmp/tmpfPUZkI.py
True
/export/gleckler1/anaconda2/envs/pmp_0717/bin/python: can't open file 'pcmdi_compute_climatologies.py': [Errno 2] No such file or directory

DRIVER IS ON OCEANONLY
/export/gleckler1/processing/metrics_package/my_test/climo_cmor/cmip_clims_param_parallel.py

doutriaux1:
@gleckler1 you need the full path to pcmdi_compute_climatology. You're right it's not really great let me push a quick fix for this.

gleckler1:
@doutriaux1 that helped but led to a new set of errors....
Executing: /export/gleckler1/anaconda2/envs/pmp_0717/bin/python /export/gleckler1/anaconda2/envs/pmp_0717/bin/pcmdi_compute_climatologies.py -p /tmp/tmpNFt7wk.py
True
  File "/export/gleckler1/anaconda2/envs/pmp_0717/bin/pcmdi_compute_climatologies.py", line 147, in <module>
    filename = glob.glob(filename_in())[0]

Not obvious...
[cmip_clims_param_parallel.py.txt](https://github.com/PCMDI/pcmdi_metrics/files/2215222/cmip_clims_param_parallel.py.txt)

doutriaux1:
@gleckler1 ok give it a try now

doutriaux1:
why do you need "Files" it's an issue with your parameter file, I changed pcmdi_compute_climatologies, we do not need files anymore, start from the parameter I updated in the example directory. I don't know what else you're doing here.

doutriaux1:
hum let me see, might be me

gleckler1:
Could be both of us.

doutriaux1:
@gleckler1 ok I had not pushed the updated param file.... Please try again using the one I just pushed up.

gleckler1:
@doutriaux1 thks... will try after telecon... or maybe during ;-)

gleckler1:
@doutriaux1  OK, now working but adding this in the parameter file makes it choke...
f  = open(sys.prefix + "/share/pmp/cmip_model_list.json")
possible_mod_dic = json.load(f)
modnames = possible_mod_dic[MIP][exp]

doutriaux1:
@gleckler1 I can see that, the paramerter thing needs some sanitation. @zshaheen when do think you'll be done with the part that only brings in "allowed" parameters in, rather than all of them, that would solve this.

doutriaux1:
@zshaheen any progress on your end with CDP, it's holding us here. Thanks.

zshaheen:
@doutriaux1 This is regarding making sure that only the parameters used are the ones defined in the parser correct? If so, I'll get to this now.

doutriaux1:
@zshaheen that would be great, I think it would help with this issue, plus we really want to only bring what was asked for not anything else. Thanks!

zshaheen:
@doutriaux1 The lastest commit for CDP, https://github.com/CDAT/cdp/commit/fc63c1053237845f5639a7a80535958f1e5d7fa4, implements this.

doutriaux1:
@zshaheen awesome will try it out shortly! Thanks a zillion!

zshaheen:
@doutriaux1 One point though. To do this, you need to have `parser.get_parameters(check_values=True)`. The default is `check_values=False`.

The default is False because the idea is that some scientist who just wants to use CDP might not be able to implement a proper command line parser. Also for other diagnostic packages, it breaks the software.

So for PMP, if we want to force this new checking, we'd just have to overwrite `PMPparser.get_parameters()` and have it call it's parent like so:

```python
class PMPMetricsParser(cdp.cdp_parser.CDPParser):
    def get_parameters(self, check_values=True, *args, **kwargs):
        return super(PMPMetricsParser, self).get_parameters(check_values, *args, **kwargs)
```

doutriaux1:
@zshaheen I can live with this but I really think the default should be True. Because that's the expected behavior really if you do `--crap=some_crap` ArgumentParser will complain it does not know about `crap` so we would stay consistent.

gleckler1:
@doutriaux1  Need to start using this code... all that is needed is enabling A.results_dir (below) to be a genutil string set in the param file...

line 530: 
nm = os.path.join(A.results_dir, "{}_PMP_{}_{}_r{}i{}p{}_{}{:02d}-{}{:02d}-clim-{}.nc".format(
v, model_id, exp, r, i, p, tc[0].year, tc[0].month, end_tc.year, end_tc.month, season))

doutriaux1:
First thing in the morning, maybe even today

C.

doutriaux1:
@gleckler1 it should be good to go now. Please note I changed the parameter `--var` to `--variable`

gleckler1:
@doutriaux1  I will test this out now.

gleckler1:
@doutriaux1  --bash option is now working well, so we can go with this for now... however, looks like all test have failed...

PR #572, opened 2018-09-19
Add coverage1

muryanto1:
add coverage & update .circleci/config.yml to run linux py3 with coverage, only do py3 if py2 passes.
update tests/test_pmp_parser.py to run when run_tests.py is run with "-c" option.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ aliases:
        PKG_NAME: "pcmdi_metrics"
        LABEL: "linatest"
     command: |
-       if [[ $CIRCLE_BRANCH != 'add_coverage' ]]; then
+       if [[ $CIRCLE_BRANCH != 'add_coverage1' ]]; then
           exit 0
        fi
        export PATH=${HOME}/project/$WORKDIR/miniconda/bin:$PATH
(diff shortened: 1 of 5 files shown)


durack1:
@muryanto1 thanks for this!!

.circleci/config.yml
```diff
@@ -64,70 +77,100 @@ aliases:
        cd conda-recipes
        rm -rf cdp
        cp -r ../../recipes/pcmdi_metrics .
-       python ./prep_for_build.py -l 1.2
-       conda build pcmdi_metrics -c conda-forge -c cdat -c pcmdi --python=27
-       conda build pcmdi_metrics -c conda-forge -c cdat -c pcmdi --python=36
-       anaconda -t $CONDA_UPLOAD_TOKEN upload -u pcmdi -l nightly $CONDA_BLD_PATH/$OS/$PKG_NAME-*tar.bz2 --force
+       python ./prep_for_build.py -l 1.1.2
```
doutriaux1: should be 1.2

tests/all_coverage.json
```diff
@@ -0,0 +1,3 @@
+{
+   "include": ["pcmdi_metrics", "cdat_info", "cdms2", "genutil", "vcs", "vtk"]
```
doutriaux1: we do not want coverage on anything but pcmdi_metrics for this repo.

doutriaux1: actually I'm not sure what this is for?

muryanto1: when we run 'pcmdi_metrics' testsuite with coverage, it will only collect coverage for 'pcmdi_metrics'. (specified in tests/coverage.json).  
In cdat nightly, where we run all test suites, I want to collect coverage for all packages...so that we can see our overall coverage...for ex: pcmdi_metrics test cases uses cdms2, genutil, vcs, etc... coverage from running all test suites will be collected....and for that I will use tests/all_coverage.json.

PR #576, opened 2018-09-24
WIP 535 pjg monsoon diurnal parameterfiles

gleckler1:
@doutriaux1  For these "examples" to be complete we need to host data needed to run them...


doutriaux1:
@gleckler1 merged master in this. Will see what breaks now, but you're up to date and compatible with master in this.

PR #601, opened 2019-04-26
tweaks to monsoon_wang

doutriaux1:
@gleckler1 @lee1043 2 demos

Also some majors tweaks to monsoon wang. Mostly related to templating, the way it was it was not flexible at all and tailored to our needs. Now it's using StringConstructors and params can be added to param file to accomodate for any case.


doutriaux1:
@gleckler1 @lee1043 also please look at demos text and add to it to explain some of the science behind the two driver we run

PR #615, opened 2019-07-25
added files

gleckler1:
@doutriaux1 added set of jsons


doutriaux1:
These really belong on the website, but since the web is generated from these I guess it's ok.  BTW these have `yaml` in lower case. Do you want to regenrate them with the new `YAML` (upper case)

PR #623, opened 2019-10-22
Modes of variability (branch: 377 variability mode nao)

lee1043:
Work in progress for merging the modes variability code to the master. Please do not merge yet. The PR is not ready for merging until told by myself.


lee1043:
@muryanto1 I think the code is ready for merging, could you please take a look when you have time? Thank you very much!

lee1043:
@muryanto1 thank you for reviewing the PR. 
- doc directory is for parameter files that serve as a namelist example, collecting user given options such as directory to find input data. Thoes parameter codes are not running by themselves but always working with *py in scripts directory. 
- I agree directory name could be better name than 'doc'. But at the moment what I need to be done is placing code in the master branch for PCMDI SFA proposal, so I will think about it later.

lee1043:
@gleckler1  MoV code is in master now. As we discussed earlier we will need to work on documentation, usage example, and test suites.

PR #634, opened 2020-01-23
Conda forge pinning

muryanto1:
updated .circleci/config.yml to use the new conda-recipes' build_tools/conda_build.py which built the package using 'conda build' and install the built package using --use-local in the test environment. 
conda-recipes' build_tools/conda_build.py also builds using conda-forge-pinning.
This way, we are really testing the package which is going to be put in pcmdi/label/nightly (if tests pass).
With this PR, you will see coverage went down to from 77% to 59%. 
In the past, we ran tests with  --coverage-from-egg option which does not work anymore since we built using 'conda build', and the coverage were including coverage from the test code itself which ofcourse are 100% (since we ran all testcases), as a result, you see a higher coverage number (77%) which I think can be misleading, since we really want to see the coverage on the code itself.

I also updated the name of job in the circleci workflow, instead of linux_pmp_py3, I renamed it to linux_pmp_py37.
You can ignore the linux_pmp_py3 and macos_pmp_py3, those jobs do not exist anymore.


durack1:
@lee1043 @gleckler1 I will leave the review to you guys

lee1043:
@muryanto1 one question; any change if this change make the CircleCI indication as "Failed" because of dropped coverage percentage?

muryanto1:
@lee1043 the circleci ci "linux_pmp_py3" and "macos_pmp_py3" failures has nothing to do with coverage. They showed up as failed because those 2 jobs do not exist anymore. Once this PR merges, CircleCI should not show those 2 jobs.

lee1043:
@muryanto1 good to know, thanks for clarifying.

muryanto1:
@lee1043 @gleckler1 Jiwoo gave me a go to merge this PR.

PR #637, opened 2020-06-04
Libnetcdf variants

muryanto1:
@lee1043 Could you take a look at the test failures?
https://circleci.com/gh/PCMDI/pcmdi_metrics/tree/libnetcdf_variants
so this is picking up the latest nighties of cdat packages.


lee1043:
@muryanto1 I think the failed test is not relevant to this change. It may need to be addressed in separate to this. I don't think I have any problem to give it go.

PR #651, opened 2020-11-04
rename dirs

gleckler1:
@lee1043 @acordonez  I created an issue then a branch to rename the "examples" folder and one folder below it.  This may seem overkill for just changing a dir name but given our recent experience with circleci I think it is a good idea to be cautious.  

My proposal is that instead of "examples", "sample_setups" is where we maintain our internal-operations parameter files ("pcmdi_parameter_files") as well as "external-setups".   Another topic, the "jsons" folder - Jiwoo and I have previously discussed setting up a separate repo to maintain the json results across all metrics and generations of CMIP.  

Let me know if you are ok with my proposed changes or have any suggested alternatives.  If you are ok with them, one of you can merge to master.   But first,  I'm about to select "create pull request" so we'll see how circleci responds to this.   It may take many hours before it goes through all tests.


acordonez:
@gleckler1 @lee1043 The name change to "sample_setups" sounds good to me. I do want to try running the changes just to make sure nothing breaks!

gleckler1:
@acordonez  Thanks for the feedback.   Now you have me wondering if we should try using the branch a little in addition to the circleci testing.   I know there is a way to revert master back in case something goes wrong but I'm not sure how to do that.  However, a precautionary trick is to first copy master to a branch like v20201104 and then its simple to go back if something goes awry.   Anyhow, once we have the wheels greased simple changes like this should become trivial.

acordonez:
I followed Lina's instructions and had all test pass locally on this branch:
$ python setup.py install
$ python run_tests.py -H -v 2

gleckler1:
@acordonez @muryanto1 @lee1043  That is that an excellent question.   I do not know the relationship between "run_tests.py" in the PMP and what circleci does.

muryanto1:
> I followed Lina's instructions and had all test pass locally on this branch:
> $ python setup.py install
> $ python run_tests.py -H -v 2
> 
> Is this similar to what CircleCI does
---
If you look at .circleci/config.yml, in &run_tests, it calls 'make run-tests ...', and this calls the 'run-tests' target in the Makefile, which basically runs 'python run_tests.py ...'.

gleckler1:
@muryanto1 @acordonez @lee1043   If I understand this correctly it means we can quickly run tests locally on our platform via run_tests.py.   Is the primary added value of circleci to run the tests on other platforms and os versions?

muryanto1:
> @muryanto1 @acordonez @lee1043 If I understand this correctly it means we can quickly run tests locally on our platform via run_tests.py. Is the primary added value of circleci to run the tests on other platforms and os versions?

Circleci builds a conda package for pcmdi_metrics on linux and macos, and then it creates a conda env and installs that built package into the conda env, and run tests for both OS and all the python versions and the libnetcdf variants.
If all tests passed, it uploads the pcmdi_metrics package to anaconda.org/pcmdi organization and marks it 'nightly'.

lee1043 (approved):

lee1043:
@muryanto1 could you please remind me why some of circle ci tests (e.g., ci/circleci: linum_pmp_py2, 3, ...) are not proceeding? I believe it is safe to merge with the fact that test-linux-3.X-XX tests were passed, but just wanted to better understand.

muryanto1:
@gleckler1 @lee1043 Those jobs that are stalling are old circleci jobs. I am not sure why they are showing. The current jobs that matter are the ones that are named "test-linux-*, test-maco-*, build-linux and build-macos.  
so this PR is safe to merge.
We may try to recreate the webhook later to see if it will fix the problem.

gleckler1:
@muryanto1  Many thanks for the clarification!

PR #654, opened 2020-11-12
simple change

gleckler1:


acordonez:
@gleckler1 This looks good to me. Were you able to preview the changes successfully?

gleckler1:
@acordonez @lee1043 Yes, I was able to view it locally as well as via github on my personal repo.  I see that you have [a hook](https://github.com/PCMDI/pcmdi_metrics/deployments) in there already for the PMP under "Environments" on the RHS.  Looks to me that when we are ready all we need to do is point to it... or something like that.

gleckler1:
@acordonez @lee1043  This PR is for gh-pages, so there are is no visible circleci tests via GH.  But I also don't see sign of it on circleci, so it appears branches do not undergo testing until a PR is made (I think that is ok).  Ana, are you still unable to view records on circleci?  We may need to investigate what permissions you may need.

lee1043:
I agree that circle ci test may not applied for this PR (and commits), which is okay, I think.

acordonez:
@gleckler1 I still can't see "Organizational Settings" and "Plan" under CircleCI, but it looks like I have access to everything else. Not sure if I'll need access to those items for work purposes.

PR #665, opened 2020-12-16
CMEC compliant mean climate output

acordonez:
An option is added to output a cmec-compliant metrics json for the mean climate module in addition to the default metrics json.

- Create "cmec" flag in PMP parser, which can be True or False.
- Add write_cmec method to Base class.
- Add call to write_cmec method in outputmetrics.py script, which handles running metrics for mean climate.

Ultimately, the write_cmec function will be able to handle any of the metrics, but for now it is only written to convert the mean climate output to the CMEC format.

**Testing**
This has been tested in the command line using a custom parameter file with `cmec = True`:
```
python setup.py install
mean_climate_driver.py -p cmec_param.py
```
I have also run the PMP test suite after building this version and all tests passed:
`python run_tests.py -H -v 2`

**Output files**
I have attached a zip file with two output jsons (results from running the command line test above):
rlut_2.5x2.5_regrid2_linear_metrics_cmec.json - cmec output file
rlut_2.5x2.5_regrid2_linear_metrics.json - default output file
[meanclimate.zip](https://github.com/PCMDI/pcmdi_metrics/files/5699533/meanclimate.zip)

### DIFF
--- a/pcmdi_metrics/driver/outputmetrics.py
+++ b/pcmdi_metrics/driver/outputmetrics.py
@@ -173,7 +173,10 @@ class OutputMetrics(object):
         if self.check_save_test_clim(ref):
             self.output_interpolated_model_climatologies(test, test_data)
 
-        self.write_on_exit(self.parameter.cmec)
+        if hasattr(self.parameter, 'cmec'):
+            self.write_on_exit(self.parameter.cmec)
+        else:
+            self.write_on_exit(False)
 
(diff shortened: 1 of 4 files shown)


gleckler1:
@acordonez  Great!  I'll test this out this AM for a combination of uses including the use of dask and not.

gleckler1:
@acordonez  dask with other metrics (not mean climate) seems to be working.   A few more things to confirm and then we'll be good to go.

PR #671, opened 2021-01-14
664 pjg 2climcalcoptions v2

gleckler1:
@acordonez @lee1043 total redo, passing tests locally now as 664_pjg_2climcalcoptions_**v2**
We'll see how tests do here.


gleckler1:
@acordonez @lee1043  wow.  Looks like it passed.  Can one you do a quick review of this and if all looks ok go ahead and merge?  It may be possible that AO does not have sufficient privileges.  If true, another thing for us to fix.

lee1043 (approved):

lee1043 (approved):

PR #678, opened 2021-01-20
677 pjg obsdic

gleckler1:


gleckler1:
@acordonez @lee1043 I'm confused by the "conflicting file" above.   Why can't we just update/replace the version.py?  Any ideas?

acordonez:
@gleckler1 Did version.py get updated in master since you created your branch? That might cause a conflict.

gleckler1:
@acordonez  I'm not sure, and in any case don't understand the reasons or implications for doing this.  But I have cp version.py from master, pushed again, and am awaiting tests again.

gleckler1:
@acordonez @lee1043  looks like this one is good to go...

lee1043 (approved):

PR #682, opened 2021-01-28
Update mean climate notebook

acordonez:
The changes in this PR are:
1. Update the mean climate demo notebook with more examples
2. Update the download notebook to use new sample data
3. Move the parameter file generation function to a separate file
4. Update parameter files with new sample data paths
5. Add a README for the notebooks


acordonez:
@gleckler1 @lee1043 Here's my first pass at a new mean climate notebook.

gleckler1:
@acordonez I'm thinking to make a few edits to the README.md, but I'm not sure if its ok for me to directly edit the file from the pull request (looks like I have permission) or if I should just make suggestions here.

gleckler1:
@acordonez Wondering about this: 
Additionally, the user should clone the PMP to acquire the notebooks:
git clone https://github.com/PCMDI/pcmdi_metrics.git

Don't the notebooks get included when one creates a PMP env via conda?   I think the above means that a user have some basic experience with git and that they have git setup.  Unfortunately, not everyone does.   Maybe there is no way around this?

acordonez:
@gleckler1 For the README, if you're able to edit from the PR, that seems reasonable to me.

acordonez:
> @acordonez Wondering about this:
> Additionally, the user should clone the PMP to acquire the notebooks:
> git clone https://github.com/PCMDI/pcmdi_metrics.git
> 
> Don't the notebooks get included when one creates a PMP env via conda? I think the above means that a user have some basic experience with git and that they have git setup. Unfortunately, not everyone does. Maybe there is no way around this?

I am not aware of the notebooks being distributed with the conda package - I don't see any mention of them in setup.py so my initial thought is that they aren't part of the install. One alternative is that users can download the repo code as a zip file on github - does that sound like a simpler method, or does that also seem a bit complicated?

gleckler1:
@acordonez @lee1043  Thanks I understand better.  I wonder if they could be added to setup.py, or perhaps there is a good reason why they are not there?  Maybe it is reasonable to just complement them with static documentation, carry on, and see what kind of feedback we get.

lee1043:
@gleckler1 I think I would not prefer having jupyter notebook in setup.py. If I am not mistaking, setup.py is for enabling pmp functions to be imported, e.g., by doing `setup.py install`  after cloning this git repo, things like `import pcmdi_metrics` enabled. Installing from conda automatically process setup.py so user won't have to install it from source. We don't have any importable functions in jupyter notebook. So I don't think it is necessary to have those in setup.py.

gleckler1:
@lee1043 @acordonez  The advantage of making the notebooks available via the PMP env installed from conda is it could simplify accessibility.  But I appreciate that there are good reasons to not do this.  Perhaps for now we should not overly emphasize making the PMP/demos trivial to use, and just target non-beginners, e.g., some expectation that the user knows a little python, github, and is interested to try jupyter if they have not already.

acordonez:
@gleckler1 @lee1043 For now I do think distributing the notebooks via github is a good way to go. In the future we might consider how they can be boxed up with conda.

lee1043:
@gleckler1 @acordonez I agree for distributing the notebooks via github for now.

lee1043 (approved):

gleckler1:
@acordonez @lee1043 Sounds good.  Lets not merge this until AO has adopted notebooks with PG's latest tinkering with the demo data.

gleckler1:
@lee1043 @acordonez  I think this may have been merged before AO could finish updating demos with PG's latest corrections to "pmp_tutorial_files.txt".  Doubtful there is any harm done.   Not sure what the easiest path for AO is, create another branch with minor updates to demos, or to resurrect this branch.

lee1043:
I have not deleted the branch when approving this PR, so I think you can keep working on the existing branch and make PR from it once done.

acordonez:
Yes, I think I can make a new PR from this branch still.

PR #683, opened 2021-02-01
Update demos for new sample data

acordonez:
I've updated the parameter files to use the new sample data and rerun the demo notebooks with this new data.


acordonez:
@gleckler1 @lee1043 Here are the changes for the new demo data

PR #687, opened 2021-02-02
Remove comments

acordonez:
I left a couple of comments in Demo 1 from debugging that are now removed.


acordonez:
@gleckler1 Just found a couple of leftover comments in the mean climate demo, this PR is to get rid of them.

PR #698, opened 2021-02-24
Updates to monsoon notebook

acordonez:
I have updated the parameter file template to use new sample data and have added an additional example to the notebook. The regridding options have been hard-coded in the monsoon driver as regridTool="regrid2" and regridMethod="conserve".

Additionally, a "download complete" message has been added to Demo 0 to let users know that the download is finished.


acordonez:
@gleckler1 @lee1043 Here's a PR for the monsoon notebook changes (and I also threw in the Demo 0 download message)

PR #699, opened 2021-02-24
Updates to Diurnal cycle notebook

acordonez:
I have updated some of the headings and explanations in the notebook. The example cases have not been changed, but they are now run via bash cell magic instead of the subprocess module. The parameter templates were updated in a previous PR for working with the new sample data.


acordonez:
@gleckler1 @lee1043 Here are the diurnal notebook updates. I've mainly cleaned up the explanations - please let me know if I've misinterpreted anything.

PR #709, opened 2021-03-17
Climatology code can loop over variables

acordonez:
The climatology script now takes variables in a list and can substitute variable names in file names.

The climatology demo (Demo 1a) is updated to reflect these changes.


acordonez:
@gleckler1 @lee1043 I tried a quick change to add a variable loop to the climatology script. I also updated the Demo notebook (can see here: https://github.com/PCMDI/pcmdi_metrics/blob/668_ao_multiple_var_climo/doc/jupyter/Demo/Demo_1a_compute_climatologies.ipynb). Let me know what you think.

PR #712, opened 2021-03-18
Show metrics results in notebooks

acordonez:
This is a small change to the demo notebooks (mean climate, monsoon_wang, and mov). Instead of showing all the json information, only show the RESULTS field from the JSON. This will shorten the notebook and help users see the most relevant metrics data.

In the mean climate notebook, I added a "regions" field to the parameter file so that only Global is run in the default example. This also reduces the notebook length by not running multiple default regions.


acordonez:
@gleckler1 @lee1043 Not an urgent fix, but I found a couple of ways to make the notebooks a little shorter and tidier. Let me know what you think.

gleckler1:
@acordonez  Looks like this will present things much cleaner - great!   A few min ago merged a PR which means this and another required update and thus CI is rerunning.   I'll keep an eye on this.

PR #723, opened 2021-03-31
removed demo dir

gleckler1:


gleckler1:
@lee1043 @acordonez  It looks like removal of the top level demo directory did not cause a failure of CI!   But now I am wondering about the tests that are not running under CI (linux_pmp_py2, linux_pmp_py3, macos_pmp_py2, macos_pmp_py3).  These look like actual pmp computational tests, not just building envs.  Maybe CI is not running any of the computational tests?   Maybe they are only run when we execute run_tests.py -H locally?   Any thoughts on this?  If either of you think there is no problem here (removing demo dir) than you can merge this PR into master.   Ana, do you have permission to do that yet?

acordonez:
@gleckler1 It looks to me like those jobs are in the CircleCI queue now and all finished for previous merges to master - do they not finish for open PRs?

acordonez:
@gleckler1 Did some more digging - it looks like we are requiring these checks in our branch protection rules, but they might not be part of our CircleCI process anymore.

gleckler1:
@acordonez @lee1043  what are branch protection rules?   I remember Lina telling us that pmp_py2 and pmp_py3 tests were no longer executing but she was not sure why.   Maybe they are only done when we run_tests locally?   I'll try to do that in this case because its possible the parameter files in the demo dir are being used for the tests.

acordonez:
@gleckler1 You can see the rules if you look at Settings -> Branch -> Branch protection rules (edit) -> Require branches to be up to date before merging. The pipelines like "macos_pmp_py3" are all checked, but looking at .circleci I'm not sure that we actually have them programmed there any more, or where else they would be located.

lee1043:
Thanks @acordonez for reminding about the branch protection rule. @gleckler1 I don't think I have a good answer for your question, but I think as CDAT moved to not support py2, we can at least uncheck linux_pmp_py2 and macos_pmp_py2.

gleckler1:
@acordonez @lee1043  Thank you both.  Running locally all but 3 tests (test_pmp_coords.py, test_pmp_portrait, test_pmp_portrait_values_on) pass, which means that the param files in the /demo dir are not being used for the metrics testing.   So I think we can merge this one!   We should follow protocol... since I created this branch I'll let one of you merge it... unless you think we should not ;-) yet

PR #725, opened 2021-04-06
Monsoon Sperber notebook

acordonez:
Adding a demo notebook for the sperber monsoon metric. This PR also changes the name of the Monsoon (Wang) notebook so that both monsoon notebooks are part of Demo 2.


acordonez:
@gleckler1  Here's the Sperber Demo. Let me know if the notebook name changes make sense or if we should do something different.

PR #726, opened 2021-04-08
724 pjg rm testsiodir

gleckler1:


gleckler1:
@acordonez In addition to passing all CI tests I've confirmed that all tests run by run_tests.py with the exception of those relying on VCS which we are moving away from.   So I think this is good for you to merge unless you see a problem.

PR #737, opened 2021-05-06
attempted redo

gleckler1:


gleckler1:
@lee1043 @acordonez  In the past when I have had a CI problem with a PR that I did not understand I was able mysteriously make the problem go away just be recreating a new branch and trying again.  Trying that here...

lee1043:
Interesting... let's see what happens here.

gleckler1:
@lee1043 looks like this time it may have made it?

lee1043:
It is strange that only those below three tests are keep failing even after rerunning them.

test-linux-3.6-mpi_mpich
test-linux-3.6-mpi_openmpi
test-linux-3.6-nompi

Which error messages shown in common is 
```
File "/home/circleci/project/workdir/linux/miniconda/envs/test-pcmdi_metrics/lib/python3.6/site-packages/cdms2/cdmsNode.py", line 10, in <module>'
b'    import cdtime'
b'ImportError: libnetcdf.so.11: cannot open shared object file: No such file or directory'
```

I am not clear why that happens but I don't think the failure in the test caused by changes in this PR. Let me merge it to the master for now using the admin privilege.

PR #772, opened 2021-08-31
updated changes

gleckler1:
@acordonez @lee1043  add obs json files.   Any suggestions?


gleckler1:
@acordonez @lee1043  Looks like the same failures as 769_ljw_update_enso_demo    :-(

lee1043:
@acordonez and I are investigating this but it looks like cdms issue with netcdf.  I was able to reproduce the circleci error on my Mac, and trying to learn more about where the issue is coming from.

acordonez:
@gleckler1 I'll test these out with the new obs

acordonez (approved):
The new obs_info_dictionary.json worked in a test run. Just a note about the demo data - I was able to use the new mean climate demo data models, but the demo obs still have "latest" in the path instead of the date, so I pointed to the production obs instead.

lee1043:
I merge this PR to master since we know that circle ci failure is not coming from the changes in this PR.

PR #775, opened 2021-09-03
Turn output directory into variable

acordonez:
This is a small fix to the Demo 6 notebook for ENSO. In cells 12 & 14 I had hard-coded the output directory name. Now, the demo_output_directory variable is used to build the output directory name.

I only committed the change to the cells without running the notebook again. After the commit, I tested these two cells in the notebooks and they placed the output data in the correct directory.


acordonez:
@lee1043 I fixed the two cells from the ENSO notebook

lee1043 (approved):
@acordonez thanks for the fix!

PR #780, opened 2021-10-20
Add default CIA args

acordonez:
The CIA default args JSON is incorporated into the PMP shared files to assist the migration to conda-forge. 

Files:
share/setup_default_args.py: Copy of writejson.py from CIA. Edit arguments here, then run the script to generate share/DefArgsCIA.json
share/DefArgsCIA.json: The JSON containing the default arguments. This is now included in the PMP installation.
setup.py: Added share/DefArgsCIA.json to data_files. Added commands to regenerate DefArgsCIA.json when setup.py is run.
pcmdi_metrics/driver/pmp_parser.py: Get DefaArgsCIA.json file from PMP installation, not from separate CIA package install.

Testing:
To test this, you should create a new environment without the CIA (so don't use the CDAT nightly install). Then try running the mean climate driver.
```
conda create -n test1
conda activate test1
conda install -c conda-forge  python=3.6 cdms2 genutil cdutil cdp matplotlib numpy proj4 jupyterlab nb_conda nb_conda_kernels
python setup.py install
conda list (verify no cia)
```

### DIFF
--- a/pcmdi_metrics/driver/pmp_parser.py
+++ b/pcmdi_metrics/driver/pmp_parser.py
@@ -1,5 +1,6 @@
 import cdp.cdp_parser
 import pcmdi_metrics.driver.pmp_parameter
+import pkg_resources
 import os
 import sys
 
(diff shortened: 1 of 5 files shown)


lee1043 (approved):

acordonez:
@lee1043 Thanks for the testing and the flake8 fixes!

PR #781, opened 2021-10-28
Remove ymal files to start from scratch

lee1043:


acordonez:
@lee1043 had another thought - do we need the Makefile anymore if .circleci is gone?

lee1043:
@acordonez Great point. I do not have good understanding of purpose of the Makefile. If it is certain that the Makefile has been only used with circle ci, I agree we can remove it. Do you know how Makefile have been used in the workflow?

acordonez:
@lee1043  I searched for each of the targets in the Makefile (e.g. conda-rerender) in the repo and many of them don't show up outside the Makefile. The ones that do only show up in .circleci/config.yml. The CircleCI config also looks like the only place we use "make" commands in the repo. Based on that info, it looks like the Makefile is only used for the CircleCI workflows.

lee1043:
@acordonez thanks for investigating. In this PR, I removed circle ci related files and added github action workflow initiative. I agree it may be easier if we merge this to master before we ask Tom and Jason to step in. I will merge it when you approve the changes.

acordonez:
@lee1043 I don't see anything else that looks like it should be deleted.

acordonez (approved):
LGTM

PR #786, opened 2021-11-03
Remove vcs dependency

lee1043:


lee1043 (commented):
@tomvothecoder would you mind taking a quick look if I made [this change](https://github.com/PCMDI/pcmdi_metrics/pull/786/commits/acd0b87c865e79902a72fb189dbee6d0f3d19efe) properly? I wanted to add another directory where legacy code stored to be excluded from whatever test..

tomvothecoder:
@lee1043 That looks good to me. I would try activating the dev env and running the command `pytest` to see if it properly ignores the directory. The files should not appear in the coverage report console output.

lee1043:
Removing `mypy` sounds reasonable at this moment because most of PMP code has been wrote with dynamic typing, instead of static typing as checked by `mypy`.

lee1043:
Closed this PR to start with latest codes that are pre-commit compliant as result of PR https://github.com/PCMDI/pcmdi_metrics/pull/793.

PR #788, opened 2021-11-03
Github action setup initiative 

lee1043:
Copied from xcdat repo

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,7 @@ repos:
           - id: flake8
             args: ["--config=setup.cfg"]
             additional_dependencies: [flake8-isort]
+            exclude: ^sample_setups/jsons/
 
     - repo: https://github.com/pre-commit/mirrors-mypy
       rev: v0.910
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,36 @@
+exclude: "docs|node_modules|migrations|.git|.tox"
+default_stages: [commit]
+fail_fast: true
+
+repos:
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v3.4.0
+      hooks:
+          - id: trailing-whitespace
+          - id: end-of-file-fixer
+          - id: check-yaml
+
+    - repo: https://github.com/psf/black
+      rev: 21.9b0
+      hooks:
+          - id: black
+
+    - repo: https://github.com/timothycrosley/isort
+      rev: 5.9.3
+      hooks:
+          - id: isort
+
+    # Need to use flake8 GitHub mirror due to CentOS git issue with GitLab
+    # https://github.com/pre-commit/pre-commit/issues/1206
+    - repo: https://github.com/pycqa/flake8
+      rev: 3.9.2
+      hooks:
+          - id: flake8
+            args: ["--config=setup.cfg"]
+            additional_dependencies: [flake8-isort]
+
+    - repo: https://github.com/pre-commit/mirrors-mypy
+      rev: v0.910
+      hooks:
+          - id: mypy
+            args: ["--config=setup.cfg"]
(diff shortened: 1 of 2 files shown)


lee1043:
I guess having this `.pre-commit-config.yaml` should help some checks, at least for flake8. I copied the file from the [xcdat repo](https://github.com/XCDAT/xcdat/blob/main/.pre-commit-config.yaml). 

@tomvothecoder @acordonez could you please see if this looks okay to you?

lee1043:
Pre-commit test is failing because it expects json files to follow python style... Not sure how to tell pre-commit check to ignore `*.json` files..

acordonez:
@lee1043 I found some info about flake8 in pre-commit hooks here: https://flake8.pycqa.org/en/latest/user/using-hooks.html. Might be able to add something like "exclude: *.json"?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    - repo: https://github.com/psf/black
+      rev: 21.9b0
+      hooks:
+          - id: black
+
+    - repo: https://github.com/timothycrosley/isort
+      rev: 5.9.3
+      hooks:
+          - id: isort
```
tomvothecoder, line 21: `black` is a code formatter and `isort` is an import statement sorter.

I'm assuming the codebase did not integrate these QA tools previously. This means both of these tools need to be run on the codebase and perform formatting as needed. Otherwise, `pre-commit` will most likely throw error(s).

There are a few options moving forward
1) Run `pre-commit` locally and format everything.
  - This will produce a large diff
2) Exclude these for now and open up another issue if you want to use these QA tools 
3) Exclude them and don't use them (I personally prefer to use them)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+          - id: flake8
+            args: ["--config=setup.cfg"]
+            additional_dependencies: [flake8-isort]
+            exclude: ^sample_setups/jsons/
+
+    - repo: https://github.com/pre-commit/mirrors-mypy
+      rev: v0.910
+      hooks:
+          - id: mypy
+            args: ["--config=setup.cfg"]
```
tomvothecoder, line 37: Same comment regarding running the tools and fixing issues.

`mypy` is an optional static type checker. It may catch some existing issues that need to be fixed.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+default_stages: [commit]
+fail_fast: true
+
+repos:
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v3.4.0
+      hooks:
+          - id: trailing-whitespace
+          - id: end-of-file-fixer
+          - id: check-yaml
```
tomvothecoder, line 11: These are general QA hooks

tomvothecoder:
> Pre-commit test is failing because it expects json files to follow python style... Not sure how to tell pre-commit check to ignore *.json files..

`pre-commit` is failing due to the Trim Trailing Whitespace hook. https://github.com/PCMDI/pcmdi_metrics/runs/4094320066?check_suite_focus=true#step:4:70
I expect the other hooks to also fail based on my review comments.

Configuring the QA tools/hooks:
* `setup.cfg` from xcdat to configure `flake8`, `mypy`, and `isort`.
https://github.com/XCDAT/xcdat/blob/main/setup.cfg
* `pyproject.yoml` from xcdat to configure `black` (doesn't support `setup.cfg`).
https://github.com/XCDAT/xcdat/blob/main/pyproject.toml

.pre-commit-config.yaml
lee1043: @tomvothecoder many thanks for the info. I think your guess is correct, that such tests have not been performed for the PMP codes. I think option 2 might be the  most efficient way to move forward.

PR #808, opened 2021-11-16
Update zenodo doi badge in the front README.md

lee1043:
Update zenodo doi badge


lee1043:
Use "Cite all versions" doi badge and number

PR #822, opened 2021-11-20
Code style fix for precip variabliity codes

lee1043:
@msahn in this PR, I've separated code style change for your branch. Code style checkup and revisions were done by `pre-commt` (as documented [here](https://github.com/PCMDI/pcmdi_metrics/wiki/Tips-for-Developers#note-for-ci-process)), which now required for Github Action CI process.  

Can you please check to see if any influence to the result? I suspect not, because all changes in this PR to your branch is only for code style, but just in case it would be safer to check result consistency before merging this to your branch.

### DIFF
--- a/pcmdi_metrics/precip_variability/README.md
+++ b/pcmdi_metrics/precip_variability/README.md
@@ -1,16 +1,16 @@
 # Precip variability across time scales
 
 ## Driver code:
-./variability_across_timescales_PS_3hr_driver.py   
+./variability_across_timescales_PS_3hr_driver.py
 
 ## Parameter codes:
-./param/variability_across_timescales_PS_3hr_params_IMERG.py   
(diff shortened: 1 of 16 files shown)


msahn:
> @msahn in this PR, I've separated code style change for your branch. Code style checkup and revisions were done by `pre-commt` (as documented [here](https://github.com/PCMDI/pcmdi_metrics/wiki/Tips-for-Developers#note-for-ci-process)), which now required for Github Action CI process.
> 
> Can you please check to see if any influence to the result? I suspect not, because all changes in this PR to your branch is only for code style, but just in case it would be safer to check result consistency before merging this to your branch.

Thank you for modifying the code style. I have checked the analysis result with the modified code and found that the modification does not influence the analysis result.

msahn (approved):

lee1043:
@msahn Excellent, thank you for checking it!

PR #829, opened 2022-03-17
Add enso_metrics

acordonez:
Add enso_metrics dependency to dev environment yaml.

### DIFF
--- a/conda-env/dev.yml
+++ b/conda-env/dev.yml
@@ -18,6 +18,7 @@ dependencies:
   - cdp=1.7.0
   - eofs=1.4.0
   - seaborn=0.11.1
+  - enso_metrics
   # Testing
   # ==================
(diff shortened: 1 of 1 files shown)


gleckler1:
@acordonez Do we still rely on the CI/CD checks?

acordonez:
@gleckler1  Yes, but these are checking things like formatting mostly. The metrics testing isn't built out yet.

lee1043 (approved):

lee1043:
Thank you for the update. Can we specify version just in case?

lee1043:
@gleckler1 yes, but not through circle ci. The CI/CD process was switched to use Github Action, and at the moment, mostly about code styling.

acordonez:
@lee1043 Added and tested the version.

PR #832, opened 2022-03-30
Update enso notebook

lee1043:
ENSO demo was updated -- manual installation process for the ENSO metrics by cloning it github repository has been removed.

### DIFF
--- a/doc/jupyter/Demo/Demo_6_ENSO.ipynb
+++ b/doc/jupyter/Demo/Demo_6_ENSO.ipynb
@@ -13,19 +13,11 @@
    "id": "161973d0",
    "metadata": {},
    "source": [
-    "This notebook provides an overview of running the ENSO metrics. More information can be found in the [README]( ). Example parameter files are located in the [PMP sample setups]( ). \n",
+    "This notebook provides an overview of running the [ENSO Metrics package](https://github.com/CLIVAR-PRP/ENSO_metrics) through the PMP. More information can be found in the [README]( ). Example parameter files are located in the [PMP sample setups]( ). \n",
     "\n",
     "**Reference**: Planton, Y., E. Guilyardi, A. T. Wittenberg, J. Lee, P. J. Gleckler, T. Bayr, S. McGregor, M. J. McPhaden, S. Power, R. Roehrig, A. Voldoire, 2020: Evaluating El Niño in climate models with the CLIVAR 2020 ENSO metrics package. Bulletin of the American Meteorological Society. [doi: 10.1175/BAMS-D-19-0337.1](https://doi.org/10.1175/BAMS-D-19-0337.1)\n",
     "\n",
-    "Description for individual metrics can be found at https://github.com/CLIVAR-PRP/ENSO_metrics/wiki.\n",
-    "\n",
-    "### Requirements\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


acordonez:
@lee1043  looking at the diff it does seem that some of the metrics have changed value slightly - is that expected given the change in datasets?

doc/jupyter/Demo/Demo_6_ENSO.ipynb
```diff
@@ -801,11 +748,11 @@
       "    },\n",
       "    \"metric\": {\n",
       "      \"ERA-Interim\": {\n",
-      "        \"value\": 0.6404285425003315,\n",
```
acordonez, line 804: @lee1043 for example

acordonez: I tried a couple of tests locally - ran your new notebook using the pmp_enso_tutorial_files.v20210823.txt data and ensometrics from conda-forge, and also ran the old notebook (in main) with pmp_enso_tutorial_files.v20210823.txt and ensometrics installed locally via pip. In both cases I am seeing the same statistics as the old notebook from the diff.

lee1043: Thanks for checking. Let me try rerun the notebook in my local.

gleckler1: So, if I understand correctly, it seems as though the differences must be coming from a change in the obs.  Do I have that right?

lee1043: @gleckler1 I don't think it is caused by any OBS diff. Now the updated notebook and old notebook are using the same OBS (from the v20210823.txt), and differences in statistics are very small numbers (under point 9th digit), which I think it is negligible. I think the difference is coming from different machines were being used. 

I am attaching a file that I got from below commend,
`conda list > conda_list.txt` 
[conda_list.txt](https://github.com/PCMDI/pcmdi_metrics/files/8400335/conda_list.txt)
@acordonez could you try same on your env and see if there is any difference in versions of the PMP dependencies?

One possibility to keep the change in this PR to be more focused and traceable (most change in this PR is removing enso_metrics's manual installation part from the demo notebook), how about @acordonez uploads the new notebook ran from her local to this PR (or via new PR if that is easier)?

acordonez: @gleckler1 @lee1043 Agree that it doesn't seem to be an issue with the obs, since I am getting the same results now that I got when we did the last round of updates. Differences due to machines sounds plausible to me. There are quite a few differences between my environment and Jiwoo's also.

acordonez: I can try running the notebook and pushing the results to this PR.

acordonez:
@lee1043 I don't think I can push to this branch but I ran your notebook in another branch (832_ao_enso) and opened a PR (https://github.com/PCMDI/pcmdi_metrics/pull/836). You might be able to pull those changes here if we want to keep this PR.

lee1043:
@acordonez thank you, I am closing this PR and will merge the new PR (#836).

PR #843, opened 2022-04-21
CMIP5 and 6 comparing portrait and parallel coordinate plot

lee1043:
Jupyter notebooks for creation of portrait and parallel coordinate plots for comparing CMIP5 and CMIP6 models.


lee1043:
NOTE: Need to update to use `Metrics` class to simplify process in the notebook.

lee1043:
@acordonez I think this PR is ready for review, could you verify notebooks are working well on your end too when you have time?

acordonez:
@lee1043 The notebooks worked for me. One thing I noticed is that I had some old files in my portrait_plot/json_files that hadn't downloaded correctly in a past test (contained 404 errors). This caused a failure in the portrait plot notebook until I removed the existing json_files folder and started from scratch. Would the notebook still give the correct results if users had old versions of the same data in json_files? Should the json_files be cleaned before running each time?

acordonez (commented):
Deleted my prior comment and reposting as a review - The notebooks worked for me. One thing I noticed is that I had some old files in my portrait_plot/json_files that hadn't downloaded correctly in a past test (contained 404 errors). This caused a failure in the portrait plot notebook until I removed the existing json_files folder and started from scratch. Would the notebook still give the correct results if users had old versions of the same data in json_files? Should the json_files be cleaned before running each time?

lee1043:
@acordonez thank you for checking. I have assumed that user will run from scratch -- starting from downloading, but I see your point -- thanks for pointing it out. I might have to implement more elegant way of downloading, but I think that is for the next PR unless we find any quick solution.

acordonez (approved):

PR #845, opened 2022-04-29
Update bar chart notebook

lee1043:
Updated bar chart notebook to download raw json files, not tar


lee1043:
Only minor change made in the PR for the downloading part that has been verified already, so I'm merging it to main.

PR #849, opened 2022-05-11
Taylor Diagram notebook

lee1043:
Notebook for generating Taylor Diagrams for multiple CMIPs using PMP archived results.


acordonez (approved):
@lee1043  The notebook ran successfully for me. I wanted to flag that there were a couple of slight differences in the marker display order in Plot example 1 for me (different squares in front or back), but when I zoom in it does look like the correct data is present. The legends were exactly the same. I attached screen grabs. There were no differences in plot example 2.
Taylor_diagram_anas_run
Taylor_diagram_original
.

PR #854, opened 2022-06-09
Import Mapping function from collections.abc to comply with Python 3.10

lee1043:
To be python 3.10 compliance `Mapping` function should be imported from `collections.abc` instead `collections`, following the discussion from https://github.com/conda-forge/pcmdi_metrics-feedstock/pull/6#issuecomment-1151707431

### DIFF
--- a/pcmdi_metrics/io/base.py
+++ b/pcmdi_metrics/io/base.py
@@ -4,7 +4,8 @@ import json
 import logging
 import os
 import re
-from collections import Mapping, OrderedDict
+from collections import OrderedDict
+from collections.abc import Mapping
 
 import cdat_info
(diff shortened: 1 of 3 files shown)


acordonez (approved):
I installed this in a fresh environment with Python=3.10 and ran the test "--help" commands for the major driver scripts. It looks like this import statement is working for me now.

lee1043:
@acordonez Excellent, thank you for checking on it. Should we re-release v2.3 or do a new release as v2.3.1? I am fine with either, just wanted to check with you which would be easier in the release workflow?

lee1043:
@acordonez I just released v2.3.1, could you please update feedstock?

PR #862, opened 2022-07-21
Update download sample data script

acordonez:
Made changes to the body of download_sample_data.py to make it useful for downloading sample data in the cmec module manager. No changes were made to the generate_parameter_files function used in Demo 0. As far as I can tell, this will not affect any of the Demo notebooks or other scripts in the PMP.


lee1043 (approved):
Confirmed the change does not influence Demo 0.

PR #864, opened 2022-07-30
Add AMO for MoV

lee1043:
Add Atlantic Multidecadal Oscillation (AMO) domain and mode for variability modes metrics.


acordonez (approved):
I ran the Modes of Variability notebook and reproduced your results. Looks good to me.

lee1043:
@acordonez thanks for checking it!

PR #872, opened 2022-09-16
New capability in parallel coordinate plot

lee1043:
* Added option for fill between two lines
* Some adjustments made including default options for box and violin plots to set it as false


lee1043:
I am merging this skipping review because the PR is having minor influential to only a few plotting scripts which have been double checked.

PR #877, opened 2022-10-13
More consistency for codes style and format

lee1043:
Include 'black' in the pre-commit process.


lee1043:
The PR is for only code formatting and I confirmed that installation process is not influenced.

Issue #858, opened 2022-07-13
cmec and tests directories are being included in python package


xylar:
Because of this call to `find_packages()` without any arguments, I think you are picking up the `cmec` and `tests` directories within the python package.
https://github.com/PCMDI/pcmdi_metrics/blob/2fd4d6611ea5aef7a878690c71e1ba7a5c93509e/setup.py#L65

Changing this call to:
``` python
packages = find_packages(include='src/python')
```
would likely avoid this, though I haven't tested it myself.  (The lines immediately above this do nothing because `packages` is replaced on this line.)

Here is a screenshot from the [latest tar file](https://anaconda.org/conda-forge/pcmdi_metrics/2.3.1/download/noarch/pcmdi_metrics-2.3.1-pyhd8ed1ab_0.tar.bz2) from conda-forge:
![image](https://user-images.githubusercontent.com/4179064/178747749-9889c2bb-8e0d-4f80-95e9-96008e822d46.png)

This is not a good situation because it is unclear if `cmec` is coming from its own conda-forge package (as it should be) or from this package (not desired).

lee1043:
Hi @xylar, thank you for pointing this out. In PR #881, I am excluding `cmec` and `tests` directories from `find_packages`. 

@acordonez I don't think there is any function that needs to be imported from under the [`cmec`](https://github.com/PCMDI/pcmdi_metrics/tree/main/cmec) directory, could you please confirm on this?

@tomvothecoder I am excluding [`tests`](https://github.com/PCMDI/pcmdi_metrics/tree/main/tests) directory because I think it is yet in use. Do you think that is okay? I also wonder whether `tests` directory should be re-included when it will be used?

acordonez:
@lee1043 nothing in the cmec folder needs to be included in the pcmdi_metrics package.

xylar:
@lee1043, great, thank you!

tomvothecoder:
> @tomvothecoder I am excluding [tests](https://github.com/PCMDI/pcmdi_metrics/tree/main/tests?rgh-link-date=2022-10-19T19%3A00%3A46Z) directory because I think it is yet in use. Do you think that is okay? I also wonder whether tests directory should be re-included when it will be used?

Hey Jiwoo, yes this should be okay. As far as I'm aware, it is standard Python practice to not include the `tests` directory in the package build. This directory is used to store tests that are ran locally or in CI/CD builds.

lee1043:
[...]

PR #881, opened 2022-10-19
Update setup.py

lee1043:
In `setup.py`:
(1) Exclude cmec and tests directories from package finding, to address #858. 
(2) Remove not-in-use files (old demo, png, vcs-related). 
(3) Clean up


acordonez:
@lee1043 The installation seems to work fine using python setup.py install, but I'm not sure how to tell if the unwanted files are not being included.

lee1043:
@acordonez thanks for checking the installation. I think that is fine for now. I don't think there is a way to check what's in conda-released tar file until we tag new version and release it.

PR #885, opened 2022-10-27
Re-organize directory for mean climate metric

lee1043:


lee1043:
Progress:
- Individual metrics calculations were compared between CDAT and xCDAT, and they are consistent. 

To do:
- Calculate metrics in the workflow and compare results. 
- Check metrics for 4-d variables

PR #890, opened 2022-11-15
Increase timeout for github action build test

lee1043:
Github action test for build was failing after 5 min from starting, thus increasing it to 10 min.


lee1043:
@tomvothecoder thank you for pointing this out, it resolved the issue!

PR #898, opened 2023-02-21
Bug fix for MJO metrics

lee1043:
To prevent a warning sign like below:
`warning: failed for  GISS-E2-H r1i1p1 module 'MV2' has no attribute 'float'`


acordonez (approved):
This works for me! I was also able to reproduce the original error in the jupyter notebook, and can confirm that this fixes the error.

Issue #900, opened 2023-02-21
Error in precipitation variability demo 


lee1043:
In the [Demo](https://github.com/PCMDI/pcmdi_metrics/blob/main/doc/jupyter/Demo/Demo_7_precip_variability.ipynb)'s cell [5], I am getting the following error. 

```
Traceback (most recent call last):
  File "/Users/lee1043/mambaforge/envs/pmp_devel_20230218/bin/variability_across_timescales_PS_driver.py", line 4, in <module>
    __import__('pkg_resources').run_script('pcmdi-metrics==2.5.1', 'variability_across_timescales_PS_driver.py')
  File "/Users/lee1043/mambaforge/envs/pmp_devel_20230218/lib/python3.9/site-packages/pkg_resources/__init__.py", line 672, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/Users/lee1043/mambaforge/envs/pmp_devel_20230218/lib/python3.9/site-packages/pkg_resources/__init__.py", line 1479, in run_script
    exec(script_code, namespace, namespace)
  File "/Users/lee1043/mambaforge/envs/pmp_devel_20230218/lib/python3.9/site-packages/pcmdi_metrics-2.5.1-py3.9.egg/EGG-INFO/scripts/variability_across_timescales_PS_driver.py", line 54, in <module>
IndexError: list index out of range

---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
Cell In[5], line 1
----> 1 get_ipython().run_cell_magic('bash', '', 'variability_across_timescales_PS_driver.py -p basic_precip_variability_param.py\n')

File ~/mambaforge/envs/pmp_devel_20230218/lib/python3.9/site-packages/IPython/core/interactiveshell.py:2430, in InteractiveShell.run_cell_magic(self, magic_name, line, cell)
   2428 with self.builtin_trap:
   2429     args = (magic_arg_s, cell)
-> 2430     result = fn(*args, **kwargs)
   2432 # The code below prevents the output from being displayed
   2433 # when using magics with decodator @output_can_be_silenced
   2434 # when the last Python token in the expression is a ';'.
   2435 if getattr(fn, magic.MAGIC_OUTPUT_CAN_BE_SILENCED, False):

[...]

PR #901, opened 2023-02-23
Update precip variability demo

msahn:
To fix a bug https://github.com/PCMDI/pcmdi_metrics/issues/900#issue-1594237952, I have updated precip variability demo and driver files.


lee1043:
Confirmed that the demo is running without an error.

lee1043 (approved):

PR #906, opened 2023-02-27
Maintain documentation through main branch

lee1043:


lee1043:
I am taking back this idea. Considering additional files made for the github page, it may not a bad thing to keep that separate from the main branch.

PR #911, opened 2023-03-03
910 pjg computeclimopts

gleckler1:
@lee1043 @acordonez I'm sure there is a cleaner way to do this but these options are now functional.  OK to overhaul as needed!  The first option enables removing the clim period from the file name (useful, e.g., when clim is based on the entire time series period which is already part of the filename).  The second option is to modify the default clim output list.  Parameter file I have tested this with:  https://github.com/PCMDI/obs4MIPs-cmor-tables/blob/230_clims_pjg/inputs/pcmdi/mkClims/test_obs_param.py 
You probably don't have permission to write to obs4MIPs database so if you want to test this you probably need to change output path.


acordonez:
@gleckler1 I'll run the compute_climatologies notebook just to do some regression testing for pcmdi_metrics/mean_climate/pcmdi_compute_climatologies.py.

gleckler1:
@lee1043 any hints regarding when these simple additions can be added? ;-)

lee1043:
@gleckler1 sorry for the delay. I will check this and merge it today.

lee1043 (approved):

lee1043:
@gleckler1 just merged!

gleckler1:
@lee1043  Great, thank you!  I'll take it for a spin!

PR #920, opened 2023-04-07
Mean clim patch

lee1043:
More flexibility in mean climate usage:
* If variable name is different in filename and actual data array
* Parallel coordinate plot vertical axis range -- allow median or mean as middle, allow user give specific value for max/min of vertical axes
* Add more metadata in JSON


gleckler1 (commented):
Looks good

PR #927, opened 2023-04-28
mjo analysis for highresmip

lee1043:


lee1043:
Minor change, merging it now.

PR #933, opened 2023-05-24
open xml from pmp workflow and allow wild card usage

lee1043:


lee1043:
Minor change, merging it.

PR #934, opened 2023-06-07
use xml open capability from pmp instead of xcdat

lee1043:


lee1043:
@acordonez I found this has no impact, but would you mind double checking with mean climate demo in your env?

acordonez (approved):
These changes worked for me in the mean climate notebook.

lee1043:
@acordonez thanks for checking!

PR #936, opened 2023-06-19
Add more inline documentation

lee1043:


lee1043:
Only inline comments added. Merging it to main.

PR #948, opened 2023-06-21
Precip variability add time-averaging metric

lee1043:


lee1043:
PR for parameter files update to stay up to date.

PR #964, opened 2023-07-26
Add custom obs to mean climate driver

acordonez:
Adding an option to use the custom observations catalogue to the mean climate driver.


acordonez:
Closing to create new PR with cleaner commit history.

PR #978, opened 2023-08-30
Mean climate 3d field vertical coordinate QC

lee1043:
Mean climate 3d field vertical coordinate QC
- pressure coordinate check
- unit check
- layer key check


lee1043:
@acordonez when you get a chance, could you help checking if this PR makes any affect to the mean climate demo? No affect is expected but wanted to double check.

acordonez:
@lee1043 Yes I'll test this out today

acordonez:
@lee1043 The mean climate notebook 1b still runs with no errors. I did a before/after comparison for the Example 5 which uses zg_500 and am seeing some small (hundreths to thousandths place) differences in some of the metrics, for example rmsc_xy. It sounds like that is expected from this change?

gleckler1:
@acordonez @lee1043 Acceptable discrepancies for zg I think.  You might compare with ta which in degrees C will be more intuitive.

lee1043:
@acordonez thank you for checking this. @gleckler1 good point! I don't think we have obs for ta-850 in demo dataset, but I have compared to previously calculated from the PMP output database, and the difference is pretty subtle and I am confident on the consistency.

acordonez (approved):

PR #981, opened 2023-09-12
Add region selection from file

acordonez:
This PR adds a function that takes an xarray dataset and returns the dataset selected over a region read from file. Grid cells outside this region are converted to NaN.

Geopandas read_file() is used to load the input file. While geopandas.read_file() can load many different file types, this has only been tested with a shapefile (.shp).

The regionmask library is used to extract the gridded data over the region of interest.

Example:
```
import xcdat
from pcmdi_metrics.io.region_from_file import region_from_file

# An xarray dataset to work with
ds = xc.open_dataset("my_test_dataset.nc")
fshp = "ne_110m_admin_0_countries/ne_110m_admin_0_countries.shp"
attr = "NAME_EN"
feature = "Tanzania"
ds_rg = region_from_file(ds,fshp,feature,attr)
```

### DIFF
--- a/pcmdi_metrics/io/extract_by_shp.py
+++ b/pcmdi_metrics/io/extract_by_shp.py
@@ -7,7 +7,7 @@ import sys
 import xarray as xr
 import xcdat
 
-def mask_region(data,name,coords=None,shp_path=None,column=None):
+def extract_by_shp(data,name,coords=None,shp_path=None,column=None):
     # Return data masked from coordinate list or shapefile.
     # Masks a single region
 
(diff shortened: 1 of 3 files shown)


lee1043:
@acordonez thank you for the PR. Will we need to add `geopandas` as one of dependencies in [`dev.yml`](https://github.com/PCMDI/pcmdi_metrics/blob/main/conda-env/dev.yml)?

acordonez:
@lee1043  It's probably a good idea since I'm calling it explicitly. Geopandas is already a dependency via regionmask and probably some of the other packages we use.

lee1043 (approved):

lee1043:
@acordonez thanks for clarifying. Thanks for adding this capability. Please feel free to merge it!

acordonez:
Thanks @lee1043!

PR #982, opened 2023-09-19
Improve mean climate workflow

### LINKED ISSUE #896: Improve mean climate metrics workflow
In its conversion to use xcdat, some of existing options were omitted to reduce complexity of the conversion work. Those capabilities need to be brought back once the PR #892 is merged. 

* [ ] Use model's own land sea masking if exist
* [ ] Net flux calculation code still uses cdms, need to convert to xcdat
* [ ] Parallel processing workflow using `parallel_submitter`

lee1043:
Patch for mean climate workflow improvement


lee1043:
Merging it for minor patches.

PR #983, opened 2023-09-22
Update MJO EWR calculation

lee1043:
Update MJO EWR calculation (2023-09-21)


lee1043:
Merging it for minor patch.

PR #984, opened 2023-09-25
Remove genutil from precip variability

acordonez:
This PR is to remove the genutil dependency from the Precip Variability metrics. Instead, the Python replace function is used to substitute variables into the results directory file name.

This branch has been tested using the examples from the precip variability notebook. I created an additional test case that made use of the %(output_type) and %(case_id) variables to test those substitutions.


lee1043 (approved):
Looks good to me, thank you for the change!

PR #996, opened 2023-11-07
Parallel coordinate plot advance

lee1043:
new option: use arrow to contrast two highlighted models


lee1043:
Merging it as it is for minor patch.

PR #999, opened 2023-11-08
Portrait plot annotate text color

lee1043:
Add flexibility in selecting color for annotation text in portrait plots


bosup (approved):
This newly added option is very useful, especially for sequential color maps.

Issue #1012, opened 2023-12-22
[Feature]: statistics calculation as APIs


lee1043:
Gather statistics calculations together as shared API across different metrics, so they can be reused in PMP from different metrics and also available for direct usage. This will also help PMP's xCDAT conversion.

### Describe the solution you'd like

- Place functions under pmdi_metrics/stats, make them importable from `pcmdi_metrics.stats`
- Prepare documentation for each functions so they can be easily reused

PR #1014, opened 2023-12-23
Stats function moved to be better shared across metrics

lee1043:
This PR will address #1012

In addition, io functions for xcat-based xarray dataset were added via this PR. These functions can be imported by something like below:
```python
from pcmdi_metrics.io import get_latitude, get_longitude, get_time, get_latitude_key, get_longitude_key, get_time_key, get_axis_list
```


lee1043:
@acordonez Below functions can be imported from `pcmdi_metrics.io`. Could you review it that would be useful for your development? 
```    
    get_axis_list,
    get_data_list,
    get_latitude_bounds_key,
    get_latitude_key,
    get_latitude,
    get_latitude_bounds,
    get_longitude_bounds_key,
    get_longitude_key,
    get_longitude,
    get_longitude_bounds,
    get_time,
    get_time_bounds,
    get_time_bounds_key,
    get_time_key,
    select_subset,
```

acordonez:
> @acordonez Below functions can be imported from `pcmdi_metrics.io`. Could you review it that would be useful for your development?
> 
> ```
>     get_axis_list,
>     get_data_list,
>     get_latitude_bounds_key,
>     get_latitude_key,
>     get_latitude,
>     get_latitude_bounds,
>     get_longitude_bounds_key,
>     get_longitude_key,
>     get_longitude,
>     get_longitude_bounds,
>     get_time,
>     get_time_bounds,
>     get_time_bounds_key,
>     get_time_key,
>     select_subset,
> ```

@lee1043 Thanks for the links. It looks like these are getting implemented in the mean climate metrics - would it also be helpful for me to run those notebooks for a test?

lee1043:
@acordonez statistical computation functions were relocated to under stats directory from under mean_climate directory. Sorry for mixing two different subjects in one PR. Yes, checking mean_climate notebook would be very helpful, thank you for suggesting it!

acordonez:
@lee1043 This comment probably doesn't matter for regular lat/lon grids, if that's what we're targeting.

With the sea ice data, I'm still having some struggles with the get_latitude_key() and get_longitude_key() functions. For the dataset shown below, they'll return the 'i' and 'j' variables, which are not technically latitude and longitude even though they are the X and Y axis coordinates. 

Screenshot 2024-01-10 at 11 38 43 AM

acordonez:
@lee1043 For the rest of the functions, they are working as expected. I also ran the mean climate notebook and the results look similar to past runs.

lee1043:
@acordonez thank you for checking this. I think I would merge this PR first and handle the sea-ice data issue in separate. Would you approve the PR if you are okay with that? If you could open an issue and point to me the example dataset (if that is different to one used previously -- I think the other one had ni and nj, so this might be a different one?) that would be greatly helpful.

acordonez (approved):

PR #1020, opened 2024-01-11
Update MoV code to use xCDAT

lee1043:
This is a draft PR, opened for tracking changes.


lee1043:
Progress note: 
- Reproduced demo notebook with new code, and results are consistent. 
- Rigorous test: code conducted for all models, and found calendar related issue
- Next step: fix the calendar related issue

lee1043:
Progress note:
Rigor test -- inconsistency found for PDO
- land masking error for a few models -- found data error (lon inconsistent between sftlf and ts) -- workaround applied
- error in pcor (stat larger than 1) -- need to investigate

lee1043:
Progress note: the error in pcor that became over 1 is caused by doing regrid2 for field filled NaN (e.g., landmask) which results field filled with zero. 

Related xcdat issue: https://github.com/xCDAT/xcdat/issues/580
Resolved by: https://github.com/xCDAT/xcdat/pull/533

After updating xcdat, pcor now look more reasonable in a simple case test. 

Solution for now: keep xcdat version as its latest main branch
Permanent solution: once xcdat new version releases, update pmp dependency version in [dev.yml](https://github.com/PCMDI/pcmdi_metrics/blob/main/conda-env/dev.yml) accordingly.

lee1043:
Progress note: essential dependency update: xcdat, 0.6.1 to 0.7.0, due to update of regrid2 for its nan value handling

lee1043:
Bug in centered RMSE calcaulation (left) is resolved (right).

Screenshot 2024-04-22 at 2 04 03 PM

lee1043:
Progress note: next component needed is custom season capability

lee1043:
@acordonez when you have a chance could you please check if the demo4 notebook is working fine with this PR?

acordonez:
@lee1043 The demo notebook ran successfully. Is it correct that some figures were removed from Section 3.3.2?

lee1043:
> @lee1043 The demo notebook ran successfully. Is it correct that some figures were removed from Section 3.3.2?

@acordonez thanks for checking. Yes, that is correct. For AMO I removed monthly example and kept only yearly example which likely be more interest.

acordonez (approved):

lee1043:
@acordonez thanks for reviewing. I also asked Alex (LANL) and Shiheng for testing this code for their analysis. I am holding merge for their result.

lee1043:
Test results from Alex: [test_AJ_LANL_20240427.pdf](https://github.com/PCMDI/pcmdi_metrics/files/15141053/test_AJ_LANL_20240427.pdf)

My comment: Seeing consistent results between the old and new codes is great, although the sign reversion of the EOF3. Unless that reversion would significantly impact your analysis, I don’t worry too much about that as the EOF signs are determined arbitrarily even with a very subtle difference in the input. I suspect the change in building block (CDMS to xCDAT) may have resulted difference in higher decimal numbers. Relevant discussion can be found [here](https://www.ncl.ucar.edu/Support/talk_archives/2009/2015.html).

pcmdi_metrics/variability_mode/variability_modes_driver.py
```diff
@@ -564,9 +564,7 @@
         # try:
         if 1:
             if realization == "*":
-                run = re.split("[._]", (model_path.split("/")[-1]).split("."))[
-                    run_in_modpath
-                ]
+                run = re.split("[._]", model_path.split("/")[-1])[run_in_modpath]
```
lee1043, line 567: Shiheng found bug in this line when file is named like “file_name_1.nc”. This fix should work well either cases like file is named like “file_name_1.nc” or “file.name.1.nc”

lee1043: During the mean time, his comment was "I’ve checked the first EOF mode of PDO. The result looks good to me."

PR #1023, opened 2024-01-24
environment.yml for binderhub on LLNL nimbus

lee1043:
To enable PMP usage in the BinderHub of the LLNL's Nimbus machine, environment.yml is needed. It has the same content as the conda-env/dev.yml, thus instead of duplicating, symlink is added.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,1 @@
+./conda-env/dev.yml


lee1043:
@jasonb5 would this work for the binderhub?

Issue #405, opened 2016-09-07
Sector scale sea-ice (Ivanova et al., 2016)


gleckler1:

gleckler1:
PG to work with DI on this.

lee1043:
This project will be revisited for HiLAT-PCMDI collaboration once it is established.

lee1043:
The branch `405 pjg ivanova sic` has not been touched in a while. 

To get it back on track:
- [  ] sync the directory structure with the current PMP
- [  ] set up documentation
- [  ] (this might be later) convert the code to work with xCDAT

durack1:
@lee1043 @acordonez @gleckler1 I was having a think about the notebook, and thought I would drop my notes down here. These are just ideas, there is no expectation that this would be completed for the CAMAS workshop, but it could be if there is general appeal.

Does it make sense to have a PMP/PCMDI logo/graphic somewhere up the top of the notebook? (Can you even do that). Just thinking it might be a good template to develop, which means as these pages are being looked at, the project wins - will also be a win for a screenshot to advertise this activity to Renu. If yes, then an ESGF and potentially DOE logo (ala logos on the [CMIP6_CVs](https://github.com/WCRP-CMIP/CMIP6_CVs/#acknowledgement) acknowledgements).

Also, as we have some CMIP6 results, to inject some science into the notebook, could we compare the Ivanova CMIP5 results to the limited CMIP6 results - so 3 models in the [current notebook](https://nbviewer.org/github/PCMDI/pcmdi_metrics/blob/main/doc/jupyter/Demo/Demo_9_seaIceExtent_ivanova.ipynb) draft? Not sure if there are digital files somewhere that contain the Ivanova results - in this repo?

acordonez:
@durack1 Regarding your last question, I haven't seen the original Ivanova metrics saved digitally anywhere.

durack1:
@acordonez that's a bummer. I was just peeking at Figs 7, 8, 9 [here](https://journals.ametsoc.org/view/journals/clim/29/24/jcli-d-16-0026.1.xml#d80986528e554) and it would be really cute to generate a CMIP5 ensemble mean and compare it to our 30 CMIP6 realizations in an attempt to gauge quantified model improvement.

I wonder if @gleckler1 has anything digital that he could scrounge up? I recall in a long lost PMP branch there was some remnant code, and maybe some remnant digi results

acordonez:
[...]

PR #1027, opened 2024-01-26
sea ice notebook tweaks (#405)

durack1:
Ref #405 

@acordonez @lee1043 some suggested tweaks to add to further pad out the demo sea ice notebook


lee1043 (approved):
Looks good to me! Thanks for the change.

Issue #1036, opened 2024-01-31
[Doc]: Update Contributing Guide


acordonez:
### Describe your documentation update

Add some specific guidance about contributing new metrics to the PMP.

acordonez:
Closed by #1039

PR #1037, opened 2024-01-31
Update Contributing Guide

acordonez:
I've added a section to the Contributing Guide that specifically addresses contributing new metrics. I've also made some small changes elsewhere just to reflect the PMP contributing process more accurately.

Address #1036


acordonez:
Thanks @lee1043! I'll merge this and we can make more changes later on as needed.

PR #1038, opened 2024-01-31
logo size fix in demo notebook

lee1043:


lee1043:
Simple patch. I am merging it to fast track.

PR #1040, opened 2024-01-31
Update Sea Ice param file comments

acordonez:
I added comments explaining what the observational parameters are, similar to how I explained the model parameters. These comments will be displayed in the Demo 9 notebook.


durack1 (approved):
@acordonez looks good to me!

PR #1041, opened 2024-02-01
Remove copy statements in sea ice metrics

acordonez:
Remove copy statements to reduce memory needs in sea ice metrics.

With local testing this reduced max memory usage for the Arctic section from 12 -> 8GB, so it seems to help.

Edit: The new changes I'm pushing get the max memory to about 4Gb for me on gates, on the command line. I've refactored the code a bit and the metrics calculation order has changed a bit in the multiple realizations case. I'm getting slightly different MSE for the climatology for some realizations, and it's not clear why. The difference is usually in the 0.1 or 0.01 place. The figures look very similar but you can see some small shifts in some of the realization spread. I can keep digging into this but it's going to take me at least another day to see if this is a bug or the natural result of the refactor.

Edit2: Some more explanation of the refactor. Previously I would loop through all realizations and save the total ice extent timeseries for each realization. At the end of the loop, I'd get the model mean, calculate the overall total extent and climatologies, and get the metrics. 
The new method gets the total mean extent and climatological extent for each realization, then gets the average of those quantities over all models and calculates all the metrics. For the total extent the difference is quite small, but for the climatology this seems to make more of a difference for the MSE.


acordonez:
@lee1043 @durack1 I don't see a way to test this branch in Jason's prototype PMP binderhub link, but it runs on Gates and should reduce some of the memory usage.

durack1:
@jasonb5 just looping you in on this thread, so we can try and sync concurrent changes to get to the final binder config that works seamlessly.

Any suggestions for @acordonez to get these changes merged into your working [`update-binderhub`](https://github.com/PCMDI/pcmdi_metrics/tree/update-binderhub) branch?

durack1:
@acordonez I just took a quick peek at code, and if you wedge this somewhere, you could call this to report resource usage throughout code execution:

```python
>>> import gc, resource
>>> import numpy as np
>>> def getResources():
...     memGb = "{:05.2f}".format(np.float32(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)/1.e6)
...     pyObj = "{:08d}".format(len(gc.get_objects()))
...     print("Resources:", memGb, "Gb;", pyObj, "pyObjs in memory")
... 
>>> getResources()
Resources: 00.03 Gb; 00021319 pyObjs in memory
```

I just wedged this into the driver file as a test, I can push this to your branch if you like?

acordonez:
@durack1 @lee1043 I can shave off another gig of max usage by removing the .compute() in [L397](https://github.com/PCMDI/pcmdi_metrics/blob/15fb944c4c109db656fe3280f99b71f8a7535ffc/pcmdi_metrics/sea_ice/sea_ice_driver.py#L397) but that would limit the code to only working with observations in single files. Which is fine for the demo but not in general.

acordonez:
@durack1 I'm doing something like that in my code locally, just don't want to commit it.

lee1043:
@acordonez, just curious what if use `open_mfdataset` instead of `open_dataset` at [L385](https://github.com/PCMDI/pcmdi_metrics/blob/15fb944c4c109db656fe3280f99b71f8a7535ffc/pcmdi_metrics/sea_ice/sea_ice_driver.py#L385C11-L385C23)? `open_mfdataset` should work for a single file as well. Can you remind me what `.compute` actually does?

acordonez:
Deleted my last comment about the .compute() statement, it actually doesn't seem to make a difference when looking at multiple runs.

acordonez:
@lee1043 using open_mfdataset increases my max memory use back up, to 11GB. So it's better to use open_dataset if there's only one file to open in terms of memory.

acordonez:
Masking out the eight different regions is where the memory usage jumps from 3 to 7.5GB, so even though I'm not making new copies directly the program still uses more space for that.

durack1:
> @lee1043 using open_mfdataset increases my max memory use back up, to 11GB. So it's better to use open_dataset if there's only one file to open in terms of memory.

a plus ~3Gb/~50% delta using a different function, woah, this is bad!

acordonez:
@durack1  @lee1043   I've just been testing the Arctic section of the code; but with rearranging some steps and heavily using the del statement, I've been able to get the max memory usage down to 5.8GB. That's roughly half of what it was originally. I don't know if we can get it much lower than that. I need to clean up the code before I can push it but it could be ready sometime this afternoon.

durack1:
> @durack1 @lee1043 I've just been testing the Arctic section of the code; but with rearranging some steps and heavily using the del statement, I've been able to get the max memory usage down to 5.8GB. That's roughly half of what it was originally. I don't know if we can get it much lower than that. I need to clean up the code before I can push it but it could be ready sometime this afternoon.

This sounds like great progress. If we can get that merged into `main` then hopefully @jasonb5 should be able to rebase his branch, we can test, and hopefully get this env and binder notebook example working. If we can get these changes done today, it will allow me to ping the LANL folks for them to test too

durack1:
@sashakames just looping you on this thread, as this real-world example of resource usage is probably something that would be useful in the ESGF2-US project discussions

durack1:
> @durack1 No this is 5.8 GB for just the Arctic section of the driver code, tested on the command line.

Ah ha, ok. It might be useful to run through the whole thing, so we can ascertain the MAXIMUM memory usage, and if the nimbus machine can support it, bump each binder environment (likely ~40 concurrently) from 10240 to whatever this max usage is, plus a buffer

durack1:
@acordonez just in case it's useful, I just found https://github.com/pythonprofilers/memory_profiler/ - which appears to be alive on conda-forge ([here](https://anaconda.org/conda-forge/memory_profiler/files) the `noarch` version looks like the latest release 0.61)

durack1:
@acordonez here are a couple of other tidbits:

https://github.com/pydata/xarray/issues/3200#issuecomment-1416446874 - using `xr.set_options(file_cache_maxsize=1)` ensured memory was released
https://github.com/pydata/xarray/issues/7018#issuecomment-1242793638 - using explicit dimension chunking to reduce memory usage (presumably this gets around the netcdf memory leak, e.g. `ds = xr.open_dataset(fldr_in_grib, engine="cfgrib", chunks={"latitude":875}, backend_kwargs={'indexpath': ''}))`
And another some others with no solutions https://github.com/pydata/xarray/issues/7397, https://github.com/pydata/xarray/issues/7404

acordonez:
@durack1 Thanks. I'm working on a general fix for the regions since that made a big difference for the Arctic, then I can see if any of the profilers show other spots for improvement.

durack1:
> @durack1 Thanks. I'm working on a general fix for the regions since that made a big difference for the Arctic, then I can see if any of the profilers show other spots for improvement.

Perfect! I think while it's a little painful, we're learning alot here.

The comment I made about suppressing warnings, that could also be caught quickly (e.g. [here](https://stackoverflow.com/questions/41125598/suppress-warnings-for-python-xarray)), although we'll have to be careful about using this when we could be suppressing really useful debugging info from users that are going to want to play outside of the basic notebook template
```python
import warnings
warnings.simplefilter("ignore")
```

durack1:
@acordonez just a little python hack, you can use `var()["varName"] = something*somethingElse` as a way of dynamically assigning variables (so looping through a list of variable names). This approach has saved me ALOT of code repeats in years gone by

acordonez:
@lee1043 @durack1 I've pushed my fix to lower the memory usage of the regional subsetting. This did result in some refactoring of the model loops, and there are some small changes to the resulting metrics values. See my main comment as well. I've pushed a rerun version of the notebook. 

@durack1 Adding this to the sea ice driver did not suppress warnings for me. I can try some other things later but need to run now.
```
import warnings
warnings.simplefilter("ignore")
```

durack1 (approved):
@acordonez this looks great!

lee1043:
@acordonez thank you for the update! The numbers in JSON are consistent before and after the change, they are not different at least until 3rd decimal point.

durack1:
> @acordonez thank you for the update! The numbers in JSON are consistent before and after the change, they are not different at least until 3rd decimal point.

@lee1043 I did spot a couple of changes (there are likely a couple of extras that I missed)
- 0.0878 (old) vs 0.1074 (new) - this is MIROC6-np-r1i1p1f1-OSI-SAF-monthly_clim
- 83.8321 vs 83.5771 - MIROC6-antarctic-r2i1p1f1-OSI-SAF-monthly_clim also io, ca
- 0.6767 vs 0.5282 - E3SM-1-0-sp-r1i2p2f1-OSI-SAF-monthly_clim, also arctic

acordonez:
> > @acordonez thank you for the update! The numbers in JSON are consistent before and after the change, they are not different at least until 3rd decimal point.
> 
> @lee1043 I did spot a couple of changes (there are likely a couple of extras that I missed)
> 
>     * 0.0878 (old) vs 0.1074 (new) - this is MIROC6-np-r1i1p1f1-OSI-SAF-monthly_clim
> 
>     * 83.8321 vs 83.5771 - MIROC6-antarctic-r2i1p1f1-OSI-SAF-monthly_clim also io, ca
> 
>     * 0.6767 vs 0.5282 - E3SM-1-0-sp-r1i2p2f1-OSI-SAF-monthly_clim, also arctic

I think @durack1 caught the worst changes. I edited the main comment to add more details about how the calculations changed.

durack1:
@lee1043 bummer, I just ran the code (pre this PR merge) and maxed out memory even with the 16384 MB limit, so this refactoring is required in order for the notebook to work with the update config.

Once this is merged, I'll take another try

durack1:
Also if we add
```python
import logging
logging.getLogger('xcdat').setLevel(logging.ERROR)
```
to all files calling xcdat, then we suppress those annoying warnings, voila - this cleans up the output for users alot

acordonez:
> Also if we add
> 
> ```python
> import logging
> logging.getLogger('xcdat').setLevel(logging.ERROR)
> ```
> 
> to all files calling xcdat, then we suppress those annoying warnings, voila - this cleans up the output for users a lot. It seems to work in the sector and line plotting scripts, but dropping it into `pcmdi_metrics.io.base` didn't seem to catch it for the obs files

@durack1 if you got this working on your end would you be able to open a PR for it? The suppress warnings statement has not worked for me inside the PMP code.

acordonez:
@durack1 @lee1043 Should I merge this PR now? It sounds like some other ideas have come up in the chat that aren't related to this specific fix, which we can address in other PRs.

lee1043:
@acordonez yes please. By merging this PR we can start testing the binderhub, and additional edits can be handled by a separate PR.

PR #1042, opened 2024-02-01
Use open_dataset if single file

lee1043:
Update xcdat_openxml.py to use `open_dataset` instead of `open_mfdataset` to save memory, considering discussions at https://github.com/PCMDI/pcmdi_metrics/pull/1041#issuecomment-1922035850


acordonez (approved):
I ran the mean climate notebook to test and all cases ran successfully without changing the metrics.

lee1043:
@acordonez thank you for checking on this!

PR #1045, opened 2024-02-02
demo 9b update

lee1043:
- removed """ for pip install for more seamless demo


lee1043:
Merging the PR as it is a minor patch

PR #1048, opened 2024-02-02
Remove warnings from sea ice metrics

acordonez:
Changing the log parameters for xcdat to remove warnings.


lee1043 (approved):
Looks good to me, thank you for the update!

durack1:
Ah ha @acordonez you figured it out! I had tried embedding the logging calls inside the driver, but didn't think about dropping it into __main__ - nice! Thanks

durack1 (approved):
Looks great to me too!

PR #1054, opened 2024-02-09
Demo 9b sea-ice data explore notebook update

lee1043:
- Added more data plot for visualizing time evolution
- Added table of contents to the notebook


durack1 (approved):
@lee1043 nice update - I particularly like that you have embedded section links in the table of contents, I hadn't figured out how to do that, but it's definitely useful. One addition could be to add a "table of contents" link at each section, so you can jump from halfway through to the top, and back down as you like, it's a looooong notebook (same with the demo when considering the output)

durack1:
@lee1043 looks like this works for me https://nbviewer.org/github/PCMDI/pcmdi_metrics/blob/lee1043-patch-1/doc/jupyter/Demo/Demo_9b_seaIce_data_explore.ipynb#tmse. I presume this also works for the binder jupyterhub examples?

acordonez (approved):
I ran all cells in BinderHub. I also tried out all the links to make sure they work!

lee1043:
@durack1 great suggestion. I added links to to top. This works best for nbviewer (I added that note in the notebook as well), but I am not sure for binder-hub. If you can give a try, let me know if that works. 

@acordonez thank you for checking the links. Can you help applying similar update to Demo9?

PR #1068, opened 2024-03-10
mean climate bug fix for E3SMv2.1 evaluation

lee1043:
Bounds handling adjustment applied.


lee1043:
@acordonez when you have time, could you please check if this has any affect on demo1b? I don't expect any, but just in case.

acordonez (approved):
Notebook 1b ran for me and I didn't spot any changes.

lee1043:
@acordonez thank you for checking it!

PR #1070, opened 2024-03-14
Clean up logos

lee1043:
Removed some old logo images that no longer needed.


lee1043:
@acordonez I don't think this would affect any of current working flows but just tagged you for your interest.

acordonez (approved):

PR #1073, opened 2024-03-18
MoV: enable custom season capability

lee1043:


lee1043:
Alex has confirmed this is working well for custom season JFM

PR #1075, opened 2024-03-27
Improve the parallel coordinate plot code

lee1043:
- Added parameters to enable label for violin plot in legend and to allow more of customization of the legend box
- Minor clean up for less print on screen
- Regenerated example plots using the latest code and data


kristinchang3 (approved):
Hi Jiwoo, 

This looks great! The only error I ran into was for adding the watermark to any example images: `rotation` was set to a string of '25' instead of an integer 25. I went ahead and removed the quotations. Otherwise, these look great!

lee1043:
@kristinchang3 thanks! I am merging the PR.

PR #1076, opened 2024-04-03
Update central arctic domain

acordonez:
This PR expands the central arctic domain to remove the masking for the pole hole. On closer reading of the Ivanova (2016) paper, the pole hole was only masked for the ice edge analysis but not for the total ice area analysis.


acordonez:
@lee1043 That's correct. The OSI-SAF data we're using does have data over the hole, not sure how that's accomplished.

acordonez:
@lee1043 Also looking at this code we've been including the sea ice data over the pole in the whole Arctic metrics, but not the central arctic metrics. If we do decide to mask the hole then I need to update the total Arctic region.

lee1043:
@acordonez Do you think this exclusion of pole hole can be set as optional capability with a parameter accompanying with?

lee1043 (approved):

lee1043:
Note: pole option added by #1087 

@acordonez please feel free to merge

PR #1087, opened 2024-04-29
Add pole hole option to sea ice metrics

acordonez:
This PR add a parameter to let users control the maximum latitude (exclusive) of the Central Arctic and Arctic regions. The default value is 90.1, so that all latitudes are included.


lee1043:
@acordonez Thanks for the PR. It looks good to me, please feel free to merge.

PR #1088, opened 2024-04-30
Parallel coordinate plot option added

lee1043:
parameters added for arrow and option added for ymax/ymin setting


lee1043:
I am merging it passing review because it's a minor patch

PR #1089, opened 2024-04-30
Update documentation for sea ice

acordonez:
Update documentation to reflect new features.


lee1043 (approved):
@acordonez Looks good to me! Thanks for the PR. 

Maybe not for this PR, but I wonder if there is any chance that the two JSON files could be merged into one for simplicity.

acordonez:
@lee1043  This PR is just for updating docs. The JSONs could potentially be combined although they have slightly different hierarchies, which is why they are separate for now.

PR #1093, opened 2024-05-02
Shapely patch for metrics with cdat

acordonez:
The PMP build is failing in conda-forge (https://github.com/conda-forge/pcmdi_metrics-feedstock/pull/30) due to a segmentation fault that occurs on running `driver_monsoon_sperber.py --help`.  I reproduced the issue locally, and using the faulthandler module I found that the issue is coming from the shapely library. I've added a temporary `import shapely` statement to the driver_monsoon_sperber.py code which occurs *before* any cdat packages are imported. Once these cdat packages are removed, the `import shapely` statement and related qa flags can be removed. This fix removes the segmentation fault.


acordonez:
@bosup Just tagging you here as an FYI since you're working on the xCDAT update for monsoon sperber.

acordonez:
@lee1043 Yes that is correct. So Bo's work would not need to include this import because he is removing all cdms references. 

A minor release would be perfect.

acordonez:
@lee1043 I'm going to run through the rest of the `--help` calls locally just to make sure none of the other drivers are having this issue.

acordonez:
@lee1043 I found that the precip distribution driver and enso driver would also fail with this issue, and I've added the shapely imports there as well. These imports can all be removed when cdat is removed from those drivers.

I also noticed that the monsoon_wang_driver.py is not included in setup.py. Is that new?

lee1043:
@acordonez thank you for the adds!

I don't think monsoon_wang metrics has been fully incorporated and tested with the current PMP setup (@gleckler1 please correct me if I am wrong), but @bosup will work on it with its xCDAT conversion. So, for now I think it is okay to not include `monsoon_wang_driver.py` in `setup.py`.

PR #1096, opened 2024-05-08
Handle NaN in sea ice clims

acordonez:
The model mean metrics calculation is changed so that NaNs from an individual realization are not considered. I've tested this change locally with multiple models.
This also contains a fix to pick the most recent dated version when a wildcard is used for the version when running over many CMIP6 models.


acordonez:
@lee1043 If you run the Jupyter notebook the data results should remain the same (it's possible there have been some changes to the figure scales since we last ran the notebooks, but the actual metrics shouldn't change).

lee1043 (approved):
Looks good to me, thanks for the PR!

PR #1099, opened 2024-05-16
Add parallel execution scripts for sea ice

acordonez:


acordonez:
@lee1043  I'm opening this PR for review. I've added my personal scripts for submitting parallel jobs using the PMP sea ice metrics on CMIP models. All the files needed are in sea_ice/scripts/parallel_runs.
These scripts are added to be a reference/record-keeping and are not set up for other people to run out-of-the-box. There is no need to test-run them.

lee1043 (approved):
Thanks for the PR, looks good to me.

PR #1100, opened 2024-05-31
Cloud feedbacks: updates to code backend

mzelinka:
several lines in lib/cal_CloudRadKernel_xr.py were updated to be compatible with the latest version of xCDAT


lee1043:
Progress note: code is ready for merge after the final test -- check for any changes in the metrics due to change of landmask generation. It used to use the `global_land_mask` that is installable via only `pip`, now switched to PMP's `create_land_sea_mask` function.

lee1043:
Output JSON checked, no noticeable difference was found.

lee1043 (approved):

PR #1108, opened 2024-08-22
Add tuple fix function

acordonez:
When a tuple such as ModUnitsAdjust is read from the command line, it is split by character and not usable in this format. 

For example, the tuple `(True, 'multiply', 1e-2)` gets parsed as `('(', 'T', 'r', 'u', 'e', ',', ' ', "'", 'm', 'u', 'l', 't', 'i', 'p', 'l', 'y', "'", ',', ' ', '1', 'e', '-', '2', ')')`.

This function re-formats the tuple so that it will be correctly formatted as `(True, 'multiply', 0.01)`. If the tuple is already in this format, the function does nothing.

### DIFF
--- a/pcmdi_metrics/utils/adjust_units.py
+++ b/pcmdi_metrics/utils/adjust_units.py
@@ -26,6 +26,7 @@ def adjust_units(da: xr.DataArray, adjust_tuple: tuple) -> xr.DataArray:
             da.assign_attrs(units=adjust_tuple[3])
     return da
 
+
 def fix_tuple(target: tuple) -> tuple:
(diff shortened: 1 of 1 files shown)


lee1043:
Confirmed below both demo returns the same result

Case 1:
```python
t = (True, 'multiply', 1e-2)
fix_tuple(t)
```
Case 2:
```python
t = "(True, 'multiply', 1e-2)"
fix_tuple(t)
```

Result:
```(True, 'multiply', 0.01)```

lee1043 (approved):

acordonez:
Thanks @lee1043!

PR #1109, opened 2024-08-23
Add sftof file option to sea ice

acordonez:
1. Adds support to use model sftof or sftlf file for land/sea mask. File should use 0-1 for land sea fraction but the driver will attempt to identify when the mask uses 0-100 and convert it to 0-1. There is also an option to use no land/sea mask.
2. Makes some minor fixes to figure appearance.
3. Adds support for 360 day calendar (needed for CMIP3).
4. Cleans up units adjustment tuples read from command line.
5. Add annual mean maps.
6. Updates the netcdf option: Files are written for use in generating figures, but if netcdf=False then the netcdf files are deleted at the end of the run.


acordonez:
@lee1043 This PR has several options for using a land/sea mask file or generating it with the PMP. Should there also be an option to use no land/sea mask (i.e. if the user is confident the data is 0 or nan over land and wants to avoid the possibility of the land/sea estimate affecting results)?

lee1043:
> @lee1043 This PR has several options for using a land/sea mask file or generating it with the PMP. Should there also be an option to use no land/sea mask (i.e. if the user is confident the data is 0 or nan over land and wants to avoid the possibility of the land/sea estimate affecting results)?

I think that is a good idea! Did I suppose to review this PR? Sorry if that was slipped from my to do list. Please feel free to ping me.

lee1043:
@acordonez oh, I see the PR was just converted to "ready for review" today. Sorry for the confusion. Please let me know when I can start reviewing.

acordonez:
@lee1043 Sorry about the confusion. This is ready to review now. 
I found when running the sea ice notebook that the demo E3SM-1-0 we were originally using is no longer available. For testing purposes I've been using the other models in /p/user_pub/pmp/demo/sea-ice/links_siconc/. Do we want to change the demo notebook to remove E3SM?

lee1043:
@acordonez no problem. Yes, please consider editing the sea-ice notebook. Could you remind me if we had sea-ice sample data downloadable as a part of demo 0 or as part of the sea-ice notebook?

acordonez:
> @acordonez no problem. Yes, please consider editing the sea-ice notebook. Could you remind me if we had sea-ice sample data downloadable as a part of demo 0 or as part of the sea-ice notebook?

I don't see any in my PMP data download, and I don't remember any being added.

lee1043:
@acordonez thanks for reminding. I think there is a benefit to keep demo notebook as a standalone. Could you find a proper input data and add it to demo input data, and make it downloadable via either demo 0 or sea ice notebook, and use it throughout the notebook?

lee1043:
@acordonez let me push this back to v3.8

acordonez:
@lee1043 An update. I've found that the Demo sea ice notebooks are running on gates again (the E3SM data used seems to be back online). While we should update the notebooks to use data that users can download, that seems better suited to another PR. I think this one is ready to run and review.

lee1043:
@acordonez thank you for the update. I agree, that downloadable demo data can be in another PR. I will review this next week. I think this can be released with v3.8.1, does that sound reasonable to you?

acordonez:
@lee1043 That plan sounds good to me.

lee1043:
@acordonez I updated the demo notebook to include the output directory named `plot`. After the update I found that the notebook run without any error. The PR is looking good to me. Thank you for your work!

lee1043 (approved):

PR #1112, opened 2024-09-04
Code changes to implement PSA1 and PSA2 mode varibility analysis

zhangshixuan1987:
The Pacific South America mode is represented by an eastward-propagating wave train extending from eastern Australia to Argentina, characterized by two invariant empirical orthogonal function (EOF) patterns with associated principal component (PC) time series (PSA1 and PSA2).

In this commit, we made code changes to enable the analysis of PSA1 and PSA2 mode varibility with EOF analysis in PCMDI. Specifically,

1. regions.py was revised to include regions used for EOF analysis for PSA1 and PSA2 modes. Note that the region for PSA1 and PSA2 analysis are the same as the SAM mode
2. argparse_functions.py was revised to include PSA1 and PSA2 in the default mode varibility analysis by PCMDI

3. variability_modes_driver.py was revised to include the default setup for EOF analysis of PAS1 or PSA2 in PCMDI

4. plot_map.py was revised to define the setup of mapping plots to generate the graphics of EOF patterns and teleconnections related to PSA1 and PSA2


zhangshixuan1987:
@lee1043 : Hi Jiwoo, I have merged the code changes I made for PSA1 and PSA2 analysis to the master branch. I have run a test and successfully got the expected figures and metrics. My run script and parameter setup can be found at 

- https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.szhang/pcmdi/test_psa_mode/submit_mov_historical.sh
- https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.szhang/pcmdi/test_psa_mode/param/

The results can be found at the following links:
-https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.szhang/pcmdi/test_psa_mode/diagnostic_results/variability_modes/analysis/historical/v20240731/

- https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.szhang/pcmdi/test_psa_mode/metrics_results/variability_modes/analysis/historical/v20240731/

-https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.szhang/pcmdi/test_psa_mode/graphics/variability_modes/analysis/historical/v20240731/

One thing I noticed that could be potentially improved: I see that the PCMDI will generate all figures for the maximum # of modes from EOF analysis, for modes like PDO, SAM, etc.  I don't know if this is on purpose, but it could cause confusion for the user or reader (at least to me).  One solution I think could be useful is that: after identifying the best matching mode, it may be better to clean up the figure directory and only keep the figures for the identified best matching modes.

lee1043:
@zhangshixuan1987 thank you very much for the PR, I will take a look. 

I also appreciate your suggestion and I agree it can be helpful to clean it up. The original purpose to create 3 EOF patterns were to visually inspect matching patterns because "best" matching mode can be selected differently depending on how to define the best matching. In some cases, simulated mode with max pattern correlation does not match to mode with minimum RMSE to obs. Maybe I can make it optional to generate plots for all 3 EOF patterns. Currently when parameter `plot` is set to `False`, it turns off all figures from model, which can be elaborated.

zhangshixuan1987:
@lee1043 : Hi Jiwoo, my suggestions above are only based on the diagnostic package perspective, i.e. based on the algorithms built in the package and the configuration user selected, what is the answer we can get from the package. 

However, I agree that the mode derived from the EOF can be subjective, and including all 3 EOF patterns can be useful for double-checking and comparison. For SAM or NAM modes, I think they are pretty robust to be represented by the first EOF mode, if not, there should be some problems.  For PNA, PSA1, or PSA2 mode, EOF on model simulations can be tricky sometimes as the variance explained from the second and third EOFs can be very close, and further diagnosis such as the pattern correlation with observation EOF modes may be needed to determine the correspondence of the actual modes to the EOF analysis.

lee1043 (approved):

lee1043:
Code tested with Demo 4 notebook

PR #1123, opened 2024-09-13
QC tools

lee1043:
- QC functions for monthly time axis and time bounds checking
- Add API documentation for those functions to the instruction [website](http://pcmdi.github.io/pcmdi_metrics/)

PR #1125, opened 2024-09-15
QC tool improve

lee1043:
change API return: raise error if check not pass


lee1043:
This improves functions merged by #1123. No influence to any of existing functions. Pytest unit test improved. Merging it now.

PR #1126, opened 2024-09-15
Preparation for unified mean clim workflow, and Clean ups for monsoon and sea ice

lee1043:
This PR provides update for the followings:
- Mean climate unified workflow
- Update for sea-ice metrics
- Monsoon code clean up

I know it is not a great habit to mix two different types of work in one PR -- sorry about that.


lee1043:
It is important to make sure the new code does not interfere the current working workflow. 
- [ ] demo 1a notebook runs without error
- [ ] demo 1b notebook runs without error

PR #1129, opened 2024-09-19
bug fix for wildcard use in multiple file open

lee1043:
Addresses #1130


lee1043:
I am merging this after testing it to simplify Alex's workflow to adapt the change.

PR #1133, opened 2024-09-22
Bug fix in module mean_climate_driver.py for regional mean 

zhangshixuan1987:
Changes are made to mean_climate_driver.py to call the updated version of region_subset() function. Specifically, current mean_climate_driver.py passed three input parameters in to region_subset() to mask the data for regional mean climate calculation. However, the region_subset() in pcmdi_metrics.io library required 5 input parameters, leading to the failure of mean_climate_driver.py calculation.

The reason for the problems in mean_climate_driver.py is due to the update of the region_subset() function: 

- For the old version of mean_climate_driver.py,  region_subset() was included in **default_regions_define.py** as shown below: 

```python
def region_subset(ds, regions_specs, region=None):
    """
    d: xarray.Dataset
    regions_specs: dict
    region: string
    """
```
Where the  three input parameters ds, regions_specs, and region are used to mask the input ds data array. 

- In current version of **mean_climate_driver.py**,  region_subset() was moved to **regions.py** and update to the following formula:
```python
def region_subset(
    ds: Union[xr.Dataset, xr.DataArray],
    region: str,
    data_var: str = "variable",
    regions_specs: dict = None,
    debug: bool = False,
)
```

Where the input parameters and position are changed. However, associated changes in **mean_climate_driver.py** were not made accordingly to feel the change in regional_subset function. 

In this commit, I made corrections in  **mean_climate_driver.py** to use the new region_subset() function as mentioned above. A special note for my changes, see example below: 
```python
ds_test_tmp = region_subset(
    ds_test_tmp,
    region,
    None,
    regions_specs,
)
```
the "None" is always used for the variable name as I think this would be more general as it is not clear whether "varname" or "varname_in_file" should be passed to this function.


lee1043:
Instead of passing `None`, default for the parameter `data_var` is set to `None` in the function level. Also updated docstrings for the function. Parameter order matched and clarified by explicitly assigning.

zhangshixuan1987:
@lee1043 : Hi Jiwoo, you modifications look more clearer to me.

lee1043:
Confirmed that no error introduced to demo [1a](https://github.com/PCMDI/pcmdi_metrics/blob/main/doc/jupyter/Demo/Demo_1a_compute_climatologies.ipynb) and [1b](https://github.com/PCMDI/pcmdi_metrics/blob/main/doc/jupyter/Demo/Demo_1b_mean_climate.ipynb) notebooks.

lee1043 (approved):

Issue #1128, opened 2024-09-18
[Bug]: wrong unit conversion in mean climate diagnostics 


zhangshixuan1987:
### What happened?

The library of mean climate diagnostics includes a "load_and_regrid.py" module which attempts to extract the test and reference data at the pressure level. In the following lines:

- https://github.com/PCMDI/pcmdi_metrics/blob/fd176464087460656fcd65d2bfce3997f7e90c45/pcmdi_metrics/mean_climate/lib/load_and_regrid.py#L95C1-L98C49,
- 
below code block was added: 
```
if "plev" in list(ds.coords.keys()):
            if ds.plev.units == "Pa":
                level = level * 100  # hPa to Pa
            try:
                ds = ds.sel(plev=level)
```

While in the ``mean_climate_driver.py" module, following
[code block](https://github.com/PCMDI/pcmdi_metrics/blob/fd176464087460656fcd65d2bfce3997f7e90c45/pcmdi_metrics/mean_climate/mean_climate_driver.py#L167-L170) show up: 

```
    result_dict["Variable"] = dict()
    result_dict["Variable"]["id"] = varname
    if level is not None:
        result_dict["Variable"]["level"] = level * 100  # hPa to Pa
```

This seems to be problematic, as you have converted the unit of **level** from hPa to Pa in **mean_climate_driver.py**, and then in the **load_and_regrid.py**, **level** times 100 again when the unit of **plev** in data file is "Pa".  Usually, the unit convention for **plev** should be "Pa",  although the unit of **plev** can be "hPa" or "mb" sometimes (depend on how user prepare the data).   

Overall, I think that the unit conversion in the **mean_climate_driver.py** should be removed, and it is better to let the **load_and_regrid.py** deal with the unit conversion properly. 

 

### Environment

I am using the PCMDI master branch and run it on Argon Chryslis machine.

lee1043:
Hi @zhangshixuan1987, thank you for the post. I see your point that it could be problematic particularly when the plev in the data is not [Pa]. As PMP has been developed with multi-model context, [Pa] for plev has been expected for the most of time as a result of CMOR. 

The current code works as follows:
1. in the parameter file level is expected to be given as hPa (e.g., 850)
2. level [hPa] info is fed to `load_and_regrid.py`, but in the `load_and_regrid.py`, `level` is local variable that is not returned at the end of the function. 
[...]

PR #1137, opened 2024-09-23
Remove forced unit conversion in read_json_mean_clim.py

zhangshixuan1987:
Modify read_json_mean_clim.py module to be consistent with the changes related to https://github.com/PCMDI/pcmdi_metrics/issues/1128

In previous commit related to https://github.com/PCMDI/pcmdi_metrics/issues/1128, changes in mean_climate_driver.py were made to remove the forced unit conversion for pressure level. Accordingly, the forced unit change in read_json_mean_clim.py should be removed as well. Otherwise, the Metrics () fuction to decode the metrics json files will return wrong variable names.


lee1043:
base changed to branch `zhan391` instead of `main` to simplify review process (so it can reviewed once rather than twice from 2 PRs)

PR #1139, opened 2024-09-24
Propose a change for floating number format for mean climate metrics …

zhangshixuan1987:
Current compute_metrics.py used either ".5f" or ".3f" to form the floating numbers for the output from the mean climate driver. As the magnitude of the model variable can range from 1e5 (e.g. pressure ) to 1e-3 (e.g. humidity), using the ".5f" and ".3f" can sometimes truncate the small values to zero. It seems that using a standard scientific notation can be beneficial if the user processes the data without converting the unit data ahead. In this way, it also provides possibilities for users to convert the units within the diagnostic datasets.

I submit this pull request to trigger some discussions and see if a better solution can be made to the PMP master code.


zhangshixuan1987:
An example can be seen in the PMP metrics data for zonal wind stress (tauv) below:
- https://raw.githubusercontent.com/PCMDI/pcmdi_metrics_results_archive/refs/heads/main/metrics_results/mean_climate/cmip6/historical/v20220928/tauu.cmip6.historical.regrid2.2p5x2p5.v20220928.json

You can see that the biases in many metrics results are zeros ,while the mean values of this variable are around 10^{-2}. I think for this scenario, 3 digt decimal points for biases are not enough to  provide useful information.

lee1043 (approved):

lee1043:
@zhangshixuan1987 thank you for the PR. I am merging it now.

PR #1144, opened 2024-09-25
Allow instant build and deploy of the website

lee1043:
To allow development of fast track web doc, but it might have to be retracted for future development.


lee1043:
This PR is intended to allow website generation and deployment when PR merged to main. But this can cause confusion when new APIs were not available in the previous versions. So eventually, suggestion is to control versions of website alongside the version of the package.

PR #1149, opened 2024-09-26
Improve doc and enable more functions in API doc

lee1043:
improve docstrings, enable some functions from working branch to enable their early usage, included them in api reference docs -- requires v3.6.1 release to be used.


lee1043:
@acordonez I checked this installs well on my end, but could you also check the same on your end too? I added a few more functions to API docs that is helpful for @gleckler1's obs4MIPs work, but that means we need to release new minor version update (v3.6.1) to make the docs consistent with what actually existing functions. Once this PR merges, I will work on releasing a new version.

acordonez:
@lee1043 Yes I'll give this a test today

acordonez (approved):
I installed this new branch successfully. I also tested all of the api import statements in ipython after installation and found no issues with importation (for example, `from pcmdi_metrics.utils import create_land_sea_mask`).

lee1043:
@acordonez excellent, thank you for testing it promptly!

PR #1156, opened 2024-10-02
Additional clean up

lee1043:
clean up -- move func to lib from driver


lee1043:
Confirmed no error from demo 4

PR #1166, opened 2024-10-29
mean clim patch for on-the-fly figure generations

lee1043:
bug fix and minor adjust


lee1043:
Confirmed that demo 1b notebook runs without error

PR #1167, opened 2024-10-30
Minor patch for mean clim on-the-fly plots

lee1043:
Adjust contour levels for some variables on the plot.


lee1043:
Since its for a minor patch, I might merge this PR without asking review.

PR #1209, opened 2024-12-10
MoV code clean up

lee1043:
Clean up the code
- remove unused parts
- standardize inline documentation style


lee1043:
Confirmed no influence to demo 4 notebook. Merging it as it's mostly just clean up

PR #1229, opened 2025-01-08
consider 4-d variables for plot climatology

lee1043:


lee1043:
@ShihengDuan has confirmed that the plots were generated for 3-d variables.

Issue #1234, opened 2025-01-23
[Feature]: NCA5 shape file


lee1043:
### Is your feature request related to a problem?

Host the shape file

### Describe the solution you'd like

Enable the file as a part of the PMP, which can be retrieved by the following method. 

```python
from pcmdi_metrics import resources
egg_pth = resources.resource_path()
import os
source_path = os.path.join(egg_pth, "nca5_regions.shp")
print(source_path)
print(os.path.isfile(source_path))
```
```
/Users/lee1043/mambaforge/envs/pmp_devel_20241202/share/pmp/nca5_regions.shp
True
```

PR #1235, opened 2025-01-23
NCA5 shapefile

lee1043:
Address #1234


lee1043:
@acordonez could you review if this would meet your need?

The shape file can be found as the following.

```python
from pcmdi_metrics import resources
egg_pth = resources.resource_path()
import os
source_path = os.path.join(egg_pth, "nca5_regions.shp")
print(source_path)
print(os.path.isfile(source_path))
```

acordonez:
@lee1043 I just need to get Paul to confirm if this is the exact shapefile we want, but otherwise this has worked great! I can get the shapefile path and load the file into a geopandas dataframe and regionmask.

lee1043:
@acordonez do we still need this PR, or was this replaced by #1244?

PR #1240, opened 2025-02-04
Sample setup files for E3SM zppy workflow

lee1043:
Added sample parameter and catalogue files used for E3SM zppy workflow, which were shared by @zhangshixuan1987.


lee1043:
Merging it as it is a simple addition to the repo.

PR #1246, opened 2025-02-06
Create Database API

kristinchang3:
Creating an API to retrieve model metrics from the PMP metrics results archive database. This will allow for more efficient access to JSON data to easier construct lists of available models, metrics, etc. for the PMP viewer and other PMP graphics.


lee1043 (approved):
Confirmed the new APIs are working well via the newly added draft notebook.

PR #1255, opened 2025-02-27
Adding parameter to database API

kristinchang3:
Added parameter for retrieving model list info only


lee1043 (approved):

lee1043:
@kristinchang3 thanks for the PR. It looks good to me and confirmed the speed up.

PR #1257, opened 2025-03-07
Remove dimensions key from CMEC JSON

acordonez:
This PR drops the "dimensions" key from the "DIMENSIONS" object in the CMEC jsons.


lee1043 (approved):
Looks good to me, thank you for the quick fix!

PR #1259, opened 2025-03-07
Replace NaN with None in CMEC JSON

acordonez:
Replace instances of "NaN" in the RESULTS object with None/null. This does not check inside lists - any values inside lists will be skipped.


lee1043 (approved):
Looks good to me! No conflict occurred when running the MoV demo notebook.

PR #1277, opened 2025-04-01
monsoon wang metrics package with extended capabilities

bosup:
module added for handling multiple realizations of climate models
plotting scripts for comparing CMIP5 and CMIP6 groups added


lee1043:
Will be followed up by #1280

lee1043:
@bosup thank you for the PR!

A follow up question: it looks like SASM domain is not defined anywhere. Did you had to update [`regions.py`](https://github.com/PCMDI/pcmdi_metrics/blob/main/pcmdi_metrics/io/regions.py) to include new regions? If so, could you share the updated file as well?

lee1043:
Close this PR and follow up with https://github.com/PCMDI/pcmdi_metrics/pull/1280

PR #1295, opened 2025-05-08
Limit numpy version due to conflict in ENSO package

lee1043:


lee1043:
Just a note: numpy version was limited to `< 2.1.0` due to its conflict in ENSO package (see https://github.com/CLIVAR-PRP/ENSO_metrics/issues/91)

Issue #1303, opened 2025-07-03
[Bug]: ref_dataset_name in mean_climate_driver.py


ShihengDuan:
### What happened?

It seems in ```mean_climate_driver.py```, the variable ```ref_dataset_name``` is the dictionary of the reference dataset. The actual observation dataset name is ```ref```. 

### Minimal Complete Verifiable Example (MVCE)

```python

```

### Relevant log output

```python
--- start mean climate metrics calculation ---
varname: pr
level: None
ref: GPCP-Monthly-3-2
ref_dataset_name {'template': 'pr/gr/v20250211/pr_mon_GPCP-Monthly-3-2_RSS_gr_198301-200412_AC_v20250211_interp_2.5x2.5.nc', 'filename': 'pr_mon_GPCP-Monthly-3-2_RSS_gr_198301-200412_AC_v20250211_interp_2.5x2.5.nc'}
```

### Environment

<details>
<summary>Click to expand</summary>

- **Python version**: e.g., Python 3.9.7
- **Operating system**: e.g., Ubuntu 20.04
- **Dependencies**: List relevant dependencies and their versions, e.g., `numpy 1.23.1`, `xarray 0.21.1`

</details>

lee1043:
@ShihengDuan thank you for opening this issue and PR! I will do a few test and will merge it to the main.

PR #1304, opened 2025-07-03
fix "ref_dataset_name"

ShihengDuan:
This can fix the issue with "ref_dataset_name" https://github.com/PCMDI/pcmdi_metrics/issues/1303 with custom observations. 
I haven't check for the default obs.


lee1043:
@ShihengDuan thank you for the PR! Let me take this PR over for any further necessary change and review.

PR #1307, opened 2025-07-09
mean clim json catalogue read in

lee1043:
In case default is not specified in catalogue json (e.g., as in the new ref clim dataset catalogue json file) and only one dataset is included in the json per variable, consider it as default. 

This PR replaces #1306


lee1043:
I think this PR is also ready to review. @kristinchang3 could you review when you have time? 

Reviewing will need running demo 0, demo1a, and demo 1b notebooks after installing PMP from this branch. 

Just for a friendly reminder: [instruction for reviewing PMP PR](https://github.com/PCMDI/pcmdi_metrics/wiki/Tips-for-Developers-(2):-Test-branch)

kristinchang3:
@lee1043 Ah my mistake. I was in the docs directory instead of doc!

kristinchang3:
Hi @lee1043 I confirmed demo notebooks 0 and 1a run successfully with no errors. Demo 1b returns a file not found error in the following cell: 
image

The demo_data_tmp directory is not being created for me - unless I missed a step. When I change the --reference_data_path to 'demo_data/obs4MIPs_PCMDI_clims' the cell runs with no error. Later in the notebook, under "Extract vertical level  and customize observations" the alternate1 --reference_data_set has the data path of 'demo_data/obs4MIPs_PCMDI_clims' and runs without error. So should the top half of the notebook be using this path as well?

lee1043:
@kristinchang3 can you try something like below:

(instead of using %%bash)
```
!mean_climate_driver.py -p ... \
... \
--reference_data_path ${demo_output_directory}/obs4MIPs_PCMDI_clims' \
...
```

In this way, python variable can be used as a part of the bash commend line.

kristinchang (approved):
I would consider updating the reference data path with ${demo_output_directory} as suggested, but otherwise the demo notebooks run successfully without error.

kristinchang3 (approved):
I would consider updating the reference data path with ${demo_output_directory} as suggested, but otherwise the demo notebooks run successfully without error!

lee1043:
@kristinchang thanks for reviewing! Just merged it.

PR #1308, opened 2025-07-22
NERSC mean climate run

lee1043:


lee1043:
Minor patch. Merging it now.

PR #1314, opened 2025-08-04
Mean clim demo patch

lee1043:
Minor fix in addition to #1307


lee1043:
Merging it as it's just a minor patch on demo notebook.

PR #1315, opened 2025-08-26
Update base.py

kristinchang3:
Simplifying class Base and removing legacy cdat based libraries/functions.

### DIFF
--- a/pcmdi_metrics/io/base.py
+++ b/pcmdi_metrics/io/base.py
@@ -27,28 +27,12 @@ try:
 except Exception:
     basestring = str
 
-
-def _determine_conda():
-    """
-    Attempt to determine the path to the conda executable
-    """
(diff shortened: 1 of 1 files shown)


kristinchang3:
Hi @lee1043 - per our discussions, I have simplified the class Base in this io/base.py script. I removed functions that are not needed for json write() or write_cmec() which is the main use of this class. This removed all legacy libraries including cdat, cdutil, genutil, MV2, and cdms2. I also removed class JSONs as it is used infrequently throughout the modern version of PMP. 

Could you please run the mean climate driver in Demo notebook 1b to test? Demo notebook 0 uses the base function download_sample_data_files() as well. Thanks!

kristinchang3:
@lee1043 I am having issues merging the changes from #1316 to my branch so I think I'm going to delete this one and create a new one to keep the commit history clean and avoid causing more issues.

PR #1316, opened 2025-08-27
Allow users to specify a specific conda location

lewisjared:
Support the use of conda alternatives such as micromamba by using an environment variable.

I'm not aware of a consistent way of determining the appropriate executable in all cases.

### DIFF
--- a/pcmdi_metrics/io/base.py
+++ b/pcmdi_metrics/io/base.py
@@ -37,11 +37,28 @@ try:
 except Exception:
     basestring = str
 
-CONDA = os.environ.get("CONDA_PYTHON_EXE", "")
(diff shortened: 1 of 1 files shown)


lewisjared:
@lee1043 I think this will allow annual cycle diagnostic in the REF to pickup the correct environment. Not sure if you have a better solution

lee1043:
@lewisjared thanks for the PR!

kristinchang3 (approved):
Hi @lee1043 @lewisjared this looks good to me! I ran the mean climate driver script and compared the json conda info to the previous version to confirm the info populated is the same. Please let me know if there's anything else I can help with. Thanks!

PR #1344, opened 2026-02-03
MoV update to run smoothly with diverse datasets

lee1043:
This PR is to make the code more flexible to run it for other datasets than CMIP. In particular for COBE2. @kristinchang3 reported an issue with CODE2, which was resulted from non-standard variable name in COBE2 when it was treated like a model data. (`sst` in COBE2 rather than `ts`). The variable name `ts` was used as a part of decision to apply landmask. The change in this PR allows 'sst' in addition. Further generalization might be needed in the future.

### DIFF
--- a/pcmdi_metrics/variability_mode/lib/adjust_timeseries.py
+++ b/pcmdi_metrics/variability_mode/lib/adjust_timeseries.py
@@ -1,3 +1,4 @@
+import copy
 import xarray as xr
 
 from pcmdi_metrics.io import (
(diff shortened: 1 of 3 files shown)


kristinchang3 (changes requested):
Thanks so much @lee1043 for figuring out this fix! I just ran the new driver script successfully. I will paste the result below. I noticed that there are some extra plots that end up being produced that don't make sense. For example, there are two teleconnection plots for the obs dataset, one for sst and one for ts. The same thing occurs for the monthly plot as well. I will paste these below for your review. Is it easy to exclude those plots from being produced? 

Image

Image
Image
Image
Image

kristinchang3:
Oh good catch @lee1043 you were right - I had some images saved in the same case directory from a previous run. I just reran and everything looks good! Thanks again!

kristinchang3 (approved):

PR #1346, opened 2026-04-01
Clean up for newer xarray and pandas versions

lee1043:
- Tested Demo 1B with xarray 2026.2.0 and xcdat 0.11.2, after this fix it works without issue. 
- Taking this opportunity to clean up codes for new warning messages from newer xarray and pandas
- Re-testing other demo notebooks as well


lee1043:
@kristinchang3 thanks for the test! I added `compat='override'`, could you re-test the MOV?

kristinchang3 (approved):
Hi @lee1043 I confirmed the demo notebooks run successfully with the new change and the MOV reverted change.

PR #1368, opened 2026-04-30
Bug fix on the MOV linear regression

### LINKED ISSUE #1367: MoV linear regression conflict with newer numpy for missing value handling
https://github.com/E3SM-Project/zppy/issues/807#issuecomment-4339846024

zhangshixuan1987:
### Summary

This PR improves the robustness of `linear_regression` and EOF analysis sign-checking logic.

### Changes

- `linear_regression`: fix `LinAlgError: SVD did not converge` from newer NumPy versions by filtering NaN values before calling `np.polyfit`
- `linear_regression`: return NaN slope/intercept for grid points with invalid or insufficient samples instead of crashing
- `linear_regression`: explicitly transpose inputs to `(time, lat, lon)` before positional indexing
- `linear_regression`: use `np.asarray()` instead of `.data` to better support Dask-backed arrays
- `linear_regression`: remove unused `1e20` fill-value masking
- `arbitrary_checking`: fix SAM, PSA1, and PSA2 latitude slicing so it works with both ascending and descending latitude grids
- `gain_pseudo_pcs`: avoid in-place mutation of xarray data when reversing sign
- `gain_pcs_fraction`: use `get_time_key(pcs)` instead of `get_time_key(ds_eof_pattern)` to reliably identify the time dimension from the 1D PC array

### Motivation

The current implementation can crash during linear regression when NaNs are present in masked regions, such as land/ocean points. Newer NumPy versions may raise `LinAlgError: SVD did not converge` when invalid values are passed to `np.polyfit`.

This PR makes the regression behavior more robust by skipping invalid grid points and returning NaNs where a reliable fit cannot be computed.

### Testing

- Tested the updated linear regression workflow on data with masked/NaN grid points
- Verified that EOF sign-checking works with different latitude coordinate directions

Acknowledgement: Portions of these changes were developed with assistance from Claude and GitHub Copilot.

### DIFF
--- a/pcmdi_metrics/enso/lib/summary_plot_lib/plot.py
+++ b/pcmdi_metrics/enso/lib/summary_plot_lib/plot.py
@@ -870,21 +870,13 @@ def multiportraitplot(
             if highlight is True:
                 # find the metric color
                 if txt in met_o1 or txt + "_1" in met_o1 or txt + "_2" in met_o1:
-                    # cc = "yellowgreen" ; alpha=1.0
-                    cc = "#66c2a5"
-                    alpha = 0.45  # teal / sea-green (102, 194, 165)
+                    cc = "yellowgreen"
                 elif txt in met_o2 or txt + "_1" in met_o2 or txt + "_2" in met_o2:
-                    # cc = "plum"; alpha=1.0
-                    cc = "#8da0cb"
-                    alpha = 0.45  # muted blue / lavender-blue (141, 160, 203)
+                    cc = "plum"
                 elif txt in met_o3 or txt + "_1" in met_o3 or txt + "_2" in met_o3:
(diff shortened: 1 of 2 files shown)


zhangshixuan1987:
@lee1043 : Hi Jiwoo, this is a code change to fix the issues discussed on https://github.com/E3SM-Project/zppy/issues/807#issuecomment-4346389115.  I have done some tests following the guidance from @forsyth2, and the changes seem to resolve the issues reported there. I will ask for your review and test to confirm.

lee1043:
Confirmed that the MoV demo notebook (demo 4) run without any issue. 

@zhangshixuan1987 thank you for the PR!

lee1043 (approved):

forsyth2:
Thanks @lee1043!

PR #1374, opened 2026-05-07
Fix provenance when `conda` is not available

xylar:
E3SM-Unified now uses [pixi](https://pixi.prefix.dev/latest/) not conda for its environments.  This causes crashes in calls to pcmdi_metrics when trying to use `conda` as part of writing provenance. This PR adds support for `pixi` during provenance.

I have added a unit test to make sure provenance succeeds and shown that it does with a pixi environment I created that is identical to the `ci.yml` environment.

Finally, I removed the `defaults` channel from `ci.yml` and `dev.yml` as packages from that channel [require a license](https://www.anaconda.com/legal).

### DIFF
--- a/conda-env/ci.yml
+++ b/conda-env/ci.yml
@@ -2,7 +2,6 @@
 name: pcmdi_metrics_ci
 channels:
     - conda-forge
-    - defaults
 dependencies:
     # ==================
     # Base
--- a/conda-env/dev.yml
+++ b/conda-env/dev.yml
@@ -2,7 +2,6 @@
 name: pcmdi_metrics_dev
 channels:
     - conda-forge
-    - defaults
 dependencies:
     # ==================
     # Base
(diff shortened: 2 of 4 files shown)


lee1043:
@xylar Can this change affect conda installation, or it is just fine as all dependencies are on conda-forge?

xylar:
@lee1043, this should not affect conda installations at all.  All dependencies are on conda-forge.  You and your users do not want to get anything from defaults, as that could mean you get charged licensing fees.

xylar:
@lee1043, answering the question from #1375, those changes are not important to us at all but this one is holding up E3SM-Unified and we unfortunately do need a release with this fix pretty urgently.  If you can make a 4.0.4, that would sure help us out.

lee1043 (approved):

lee1043:
@xylar thanks for clarification. Sure, we can release 4.0.4

xylar:
Thank you very much, @lee1043!!

PR #1375, opened 2026-05-07
Fix warnings in unit tests

xylar:
This PR fixes several warnings that came up in unit tests.

There were issues with latex symbols being interpreted as escape characters, which are fixed by using r-strings.

There were two future warnings:
* `FutureWarning: cftime_range() is deprecated, please use xarray.date_range(..., use_cftime=True) instead.`
* `FutureWarning: 'M' is deprecated and will be removed in a future version. Please use 'ME' instead of 'M'`

Finally, there was this warning:
```
[WARNING]: bounds.py(create_bounds:983) >> The 'lat' coordinate variable is missing a 'units' attribute. Assuming 'units' is 'degrees_north'.
```

They are all fixed in this PR.

### DIFF
--- a/pcmdi_metrics/enso/lib/summary_plot_lib/plot.py
+++ b/pcmdi_metrics/enso/lib/summary_plot_lib/plot.py
@@ -1016,7 +1016,7 @@ def multiportraitplot(
         aspect=40,
     )
     cbar.ax.set_yticklabels(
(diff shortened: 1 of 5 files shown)


lee1043 (approved):

lee1043:
Looks good to me, thank you for the PR, @xylar!

lee1043:
@xylar how urgent this for E3SM Unified? I can either release v4.0.4 immediately or hold little bit (couple weeks?) for v4.1, depending on E3SM Unified timeline.

Issue #1380, opened 2026-05-28
[Bug]: Contour levels not centered about 0 or rounded for custom variable/levels


jsg5776:
### What happened?

When using mean_climate for a non-default variable/pressure level, contour levels were not rounded and difference plots were not centered about 0. 

### What did you expect to happen? Are there are possible answers you came across?

Recommend rounding contour levels and ensuring the minimum contour level in a difference plot is negative. 

### Minimal Complete Verifiable Example (MVCE)

```python

```

### Relevant log output

```python

```

lee1043:
Closing as #1382 was merged

PR #1382, opened 2026-05-29
Mean Climate Difference Plots

jsg5776:
Fixes for #1380 
- Renamed "levels" and "levels_diff" to include "contour" for clarity, since "levels" also refer to vertical level in the same file
- Adjusted the plotting setup so difference plots use the difference dataset when preparing colorbar settings. 
- Improved rounding behavior of the contour levels when a default variable/level is not specified.


lee1043 (approved):

lee1043:
@jsg5776 Looks good to me, thanks for the PR!

PR #1389, opened 2026-07-08
Enable public API for MJO East-West Power Ratio (EWR) calculation

lee1043:
Creation of public APIs for diagnostics will enable more flexible incorporation of PMP's diagnostics and metrics in users' workflow.


lee1043:
@kristinchang3 when you get a chance could you please review this PR? In particular, running the demo 5 notebook using codes in the main and this branches, and comparing the output files to validate if there has been any substential changes would be very helpful.

lee1043:
Note: The `mjo_metric_calc.py` code in this PR was rewritten using Claude AI tool. The code conversion was initiated using the following prompt message and debugged/revised by @lee1043.

Prompt used:
> I like the suggestion of enabling python standard API in addition to the driver scripts, so it does not disturb existing workflow if user was using the driver scripts in their workflow. I can start with one of the simplest metric: MJO east-west power ratio. The current python function is at https://github.com/PCMDI/pcmdi_metrics/blob/main/pcmdi_metrics/mjo/lib/mjo_metric_calc.py#L26, and need some modernization to enable it as a standalone API. Do you have suggestions for the improvement of the function?

kristinchang3:
Hi @lee1043, when running Demo 5 on the main branch, I encountered a small bug to due to a matplotlib udpate. Please see the PR #1391 for my suggested fix.

kristinchang3:
Hi @lee1043 - this branch needs to be updated with the changes from #1391 (now merged to main)! But after running with those changes, all of the resulting plots match except for one: cmip5_obs_historical_GPCP-IP_mjo_1998-1999_NDJFMA_cmmGrid.png . It seems there is a NaN value that may be throwing the plot off. 
Main branch: 
image

mjo_standalone_api branch:
image

lee1043:
@kristinchang3 thank you for noticing it! I have fixed the bug, and generate the plot on my end. Could you rerun to verify on your end as well?

kristinchang3 (approved):
@lee1043 thanks for the fix! That worked and it looks good to me!

lee1043:
@kristinchang3 Great, thanks!

PR #1393, opened 2026-07-17
Fix latitude ordering and land-mask naming

### LINKED ISSUE #1392: [Bug]: Monsoon Sperber Driver fails when no land fraction dataset is provided, or dataset with descending latitude dimension
### What happened?

While running the monsoon driver for a dataset with decreasing latitude dimension order and no land fraction dataset provided, the driver searched for an unknown variable in the land-fraction dataset that was created by default. Additionally, monsoon region subset failed as it assumed the latitude dimension was in ascending order.

### What did you expect to happen? Are there are possible answers you came across?

_No response_

### Minimal Complete Verifiable Example (MVCE)

```python

```

### Relevant log output

```python

```

### Anything else we need to know?

_No response_

### Environment

pcmdi_metrics_dev

jsg5776:
Fixes #1392, sorts datasets by latitude to comply with region convention and renames land-fraction dataset variable

### DIFF
--- a/pcmdi_metrics/monsoon_sperber/driver_monsoon_sperber.py
+++ b/pcmdi_metrics/monsoon_sperber/driver_monsoon_sperber.py
@@ -52,7 +52,12 @@ import xarray as xr
 import xcdat as xc
 from matplotlib import pyplot as plt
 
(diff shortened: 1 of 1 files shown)


lee1043 (approved):
Looks good to me, thanks for the fix!

PR #1397, opened 2026-07-21
Add Claude.md 

lee1043:


kristinchang3:
Hi @lee1043 - I am still reviewing this file, but one initial suggestion I have is to add a little more explicit information about the parameter files. Perhaps under the "Important Context" or "Requirements" section, something like "...existing driver scripts and corresponding parameter files (e.g., `mean_climate/param/basic_param.py`) must maintain consistent behavior for backward compatibility." 

I'm not sure if it's necessary as we are changing to more standard API format moving forward, but I thought being more explicit about how changes to the driver script need to be updated/checked against the basic_param files might guide Claude more directly.

lee1043:
@kristinchang3 thanks for the comment. I agree with you that more explicit guidance would help the tool to work in more expected way, and I just added.

kristinchang3:
Hi @lee1043 - I have been reading the Claude Code Docs (https://code.claude.com/docs/en/memory) to better familiarize myself with how the .md file works and came across a note regarding the file size: "target under 200 lines per CLAUDE.md file. Longer files consume more context and reduce adherence. If your instructions are growing large, use [path-scoped rules](https://code.claude.com/docs/en/memory#path-specific-rules) so instructions load only when Claude works with matching files."

As this file exceeds 200 lines, I think we should consider general simplification and also setting up rules as described in the note above. Perhaps as follows:

CLAUDE.md contains:
- Project Overview
- Code Modification Policy
- Workflow Compatibility
- Project Structure
- Development Best Practices
- Resources
- Important Notes

Modular rules:
- Function and Helper Creation
- Documentation Standards
- Testing
- Git workflow
- Common task (maybe each task as its own rule)

Some of the important notes are repeated throughout the previous sections (e.g., backwards compatibility, scientific integrity, new metrics) which I don't think is necessary. We could label the important notes section with **CRITICAL** instead.

If you agree, I could take a stab at revising this file and also creating separate files for the rules based on what you have here!

lee1043:
@kristinchang3 thank you for the investigation! Sure, please feel free to revise and create needed files! You are more than welcome to continue on this branch, if that works for you.

kristinchang3:
Hi @lee1043 Could you take a look at the new version of the .md and let me know your thoughts? 
I focused on concise language, removing repeated guidance, actionable guidance, and only including what is specific to PMP (as opposed to a general software development guideline which Claude is already programmed to adhere to). I think this may be a more clear starting point for any chats and then we can add some more specific instructions (as skills) for individual tasks if we find we need that.

I have your original version here for reference: 
[rough_draft.md](https://github.com/user-attachments/files/30482385/rough_draft.md)

lee1043:
Hi @kristinchang3, thanks for the update! It looks great to me! Just one comment:
- Could you additionally mention xarray under the "Key Dependencies" instruction?

kristinchang3:
Hi @lee1043 just added it!

lee1043:
Thanks, @kristinchang3, for advancing this PR! I am merging it now.

kristinchang3:
Thanks @lee1043 ! FYI I just did a test by asking Claude to review the Hadley Cell code and I feel that the suggestions for improving my code are much closer to what I want than before!

PR #1416, opened 2026-08-21
Return value improvments in extremes driver

### LINKED ISSUE #1401: [Bug]: Extremes Driver GEV/Return Period Implemenation Bugs and Optimization
### What happened?

Testing the extremes driver, noticed the following problems that could be fixed/optimized

- Noted incorrect term in the log-likelihood probability implementation for the Gumbel case
- Secondary stationary optimization even though genextreme.fit already provides the stationary MLE
- Performs the grid-cell fits serially which can be time-waste
- Filters NaN values by converting to zeros, which could be numerically valid block maxima that are then input into the GEV fit. 

### What did you expect to happen? Are there are possible answers you came across?

_No response_

### Minimal Complete Verifiable Example (MVCE)

```python

```

### Relevant log output

```python

```

### Anything else we need to know?

_No response_

### Environment

pcmdi_metrics_dev environment

jsg5776:
#1401 Improvements

- Use joblib to parallelize RV computation in space. 
- Bug fix in the Gumbel GEV case
- Normalization options in the RV function


jsg5776:
@lee1043 I tested the extremes Demo notebook to ensure compatibility

lee1043:
@jsg5776 thank you for the PR! I was able to run the Demo 8 notebook (after running demo 0 notebook) which worked without any error. I however noticed subtle differences in the metrics number compared to the [current version](https://pcmdi.github.io/pcmdi_metrics/examples/Demo_8_extremes.html). Although I don't think that is too concerning, to reassuring, do you have any interpretation for where those differences might have come from?

lee1043 (approved):

jsg5776:
@lee1043 Thank you for bringing this up. The largest numerical difference comes from the way the data are normalized. The old version normalized the input data by a global mean prior to the GEV fit, then multiplied by the same global mean to get the final return value / standard error. The new version normalizes on a _per grid point_ basis, i.e. it subtracts the mean and divides by the standard deviation of a single grid point's timeseries prior to the GEV fit. I have found that this produces more stable results, while being numerically close to our prior results in Demo 8. I have a few different normalization options now as user-input, however, they don't exactly reproduce the old results as they are all on a per-grid-point basis. If you would like, I could add an option to normalize the old way just to ensure the new methods otherwise match the old results.

lee1043:
@jsg5776 thank you for the investigation and clarification! It seems the new way of normalization makes more sense as in that way the metrics is derived for each location basis, which could be the default option. I also like the idea of adding the option for the old way so if needed that could be reproduced.

lee1043:
@jsg5776 please also include the updated demo 8 notebook in this PR

jsg5776:
Hi @lee1043. I added my changes to Demo notebook 8, which include a new example using a different normalization option. Additionally, the notebook runs each normalization method and creates a comparison plot. However, these plots do not show as much difference between methods as I might have expected, so I do not mind if we delete that section of the Demo notebook.

PR #1423, opened 2026-09-03
Lower SAM/NAM diagnostic memory footprint by 50x

lewisjared:
For the Climate-REF SAM/NAM diagnostics I was seeing a huge jump in wall time and memory usage for higher resolution runs. Up to 150 GB and 2 1/2 hours for a half-degree grid when the mean case took 8mins.
For the CMIP7 runs for UKCM2a-0-HH, it used more than 700GB of RAM.

The issue was that the Dask chunks were extremely inefficient when calculating the EOF, causing O(n^2) memory usage and lower run times. This fix just materialises the data so it uses plain numpy. The materialised data is only about 85MB at that point in time.

Runs went from 154GB, 165 min -> 2.9GB, 4 min


lee1043 (approved):

lee1043:
@lewisjared thank you for the addition!

